PR #1, opened 2024-11-06
[DSFaap-425] create service start page

DrogoNevets:
- **made start page goes no where**
- **start page**

### DIFF
--- a/src/server/home/index.njk
+++ b/src/server/home/index.njk
@@ -1,10 +1,6 @@
 {% extends 'layouts/page.njk' %}
 
 {% block content %}
-  {{ appHeading({
-    text: heading,
-    caption: "Core delivery platform Node.js Frontend Template."
-  }) }}
 
   <div class="govuk-grid-row">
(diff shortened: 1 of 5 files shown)


src/server/home/index.njk
```diff
@@ -1,56 +1,48 @@
 {% extends 'layouts/page.njk' %}
 
 {% block content %}
-  {{ appHeading({
-    text: heading,
```
hughfdjackson, line 5: We could do with updating the heading here too - to "Move animals under disease control restriction"?

src/server/home/index.njk
```diff
@@ -1,56 +1,48 @@
 {% extends 'layouts/page.njk' %}
```
hughfdjackson, line 1: I think this importing an 'about' - I think we should remove it?

hughfdjackson (changes requested):

hughfdjackson (approved):

PR #2, opened 2024-11-07
[DSFAAP-425] added app phase banner component

DrogoNevets:


src/server/common/components/phaseBanner/template.njk
```diff
@@ -0,0 +1,10 @@
+<div class="govuk-phase-banner" data-testid="app-phase-banner">
+  <p class="govuk-phase-banner__content">
+    <strong class="govuk-tag govuk-phase-banner__content__tag" data-testid="app-phase">
+      {{ params.phase }}
+    </strong>
+    <span class="govuk-phase-banner__text">
+      This is a new service – your <a class="govuk-link" href="#">feedback</a> will help us to improve it.
+    </span>
+  </p>
+</div>
```
hughfdjackson, line 10: Because we've got access to nunjucks templates from gov design system, we could replace this with: 

```
{% from "govuk/components/phase-banner/macro.njk" import govukPhaseBanner %}

{{ govukPhaseBanner({
  tag: {
    text: params.phase
  },
  html: 'This is a new service – your <a class="govuk-link" href="#">feedback</a> will help us to improve it.'
}) }}
```

I generally prefer to use the provided templates by default, because it's very easy to tell that we're following what GDS design system says (and therefore are getting the accessibility, testing, design, etc benefits).  With the HTML version, it leaves that question a bit more open.

src/server/common/components/phaseBanner/template.test.js
```diff
@@ -0,0 +1,26 @@
...
+  test('Should contain expected heading', () => {
+    expect($phaseBanner('[data-testid="app-phase"]').text().trim()).toBe(
+      'Private beta'
+    )
+  })
+})
+
+/**
+ * @import { CheerioAPI } from 'cheerio'
+ */
```
hughfdjackson, line 26: Not 100% sure what this comment is for?  Presumably it's somewhere in the template already, but would be nice to clean up.

hughfdjackson (changes requested):

src/server/common/components/phaseBanner/template.njk
DrogoNevets: its the very first line that I couldnt find

had thought it should be there but couldnt find it - updated

hughfdjackson (approved):

PR #8, opened 2024-11-11
[DSFAAP-432] added first question and store answer in cache

DrogoNevets:
- **can i push**
- **fixed test**


src/server/on-off-farm/index.njk
```diff
@@ -0,0 +1,56 @@
+{% from "govuk/components/back-link/macro.njk" import govukBackLink %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+
+{% extends 'layouts/page.njk' %}
+
+{% block metaBeforeContent %}
+  {{ govukBackLink({
+    text: "Back",
+    href: "/"
```
hughfdjackson, line 10: I think this one needs a bit of a review with Guy - I can see that the component out-of-the-box specifies a href rather than tapping into javascript history API. 

The guidance & implementation are a bit confusing: 

> Make sure the link takes users to the previous page they were on, in the state they last saw it.

does not match the offered functionality, looking at https://design-system.service.gov.uk/components/back-link/.

Unless we pick up the HTTP referrer & populate it in here?  Which would work with javascript & match the spirit of the guidance (but not the examples)

src/server/home/controller.js
```diff
@@ -6,8 +6,8 @@
 export const homeController = {
   handler(_request, h) {
     return h.view('home/index', {
-      pageTitle: 'Home',
-      heading: 'Home'
+      pageTitle: 'Apply for an animal disease movement licence',
```
hughfdjackson, line 9: This content will ultimately have to be replaced, but no arguments for now.

src/server/on-off-farm/controller.js
```diff
@@ -0,0 +1,48 @@
...
+      pageTitle: 'Are you moving the cattle on or off your farm?',
+      heading: 'Are you moving the cattle on or off your farm?',
+      onOffFarm: {
+        value: onOffFarm
+      }
+    })
+  }
+}
+
+export const onOffFarmPostController = {
```
hughfdjackson, line 20: Type declaration here?

src/server/on-off-farm/index.js
```diff
@@ -0,0 +1,36 @@
...
+ */
+export const onOffFarm = {
+  plugin: {
+    name: 'to-or-from-own-premises',
+    register(server) {
+      server.route([
+        {
+          method: 'GET',
+          path: '/to-or-from-own-premises',
+          ...onOffFarmGetController
```
hughfdjackson, line 22: I'm sure we'll figure this out later, but it feels weird to me that we're expecting some subset of properties on the route to be defined here, and some in the control itself.  Leaves open ambiguity for where a route option would be defined. 

It'd feel more natural to have the definition file be:

```javascript
const onOffFarmGetHandler = (req, h) => {
  //...
}
```

And then have this file: 
```javascript
server.route([
  {
    method: 'GET',
    path: '/to-or-from-own-premises',
    handler: onOffFarmGetHandler
}
```

This is a comment to get your thoughts rather than a request for change.

src/server/on-off-farm/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * Sets up the routes used in the home page.
+ * These routes are registered in src/server/router.js.
+ */
+
+/**
+ * @satisfies {ServerRegisterPluginObject<void>}
+ */
+export const onOffFarm = {
```
hughfdjackson, line 14: There may be a reason for this, but this could be easily cut down to just a list of routes, and then in server.js we can call `service.routes([...onOffFarmRoutes])`.

This one's a comment rather than a request for change.

src/server/on-off-farm/controller.js
```diff
@@ -0,0 +1,48 @@
+/**
+ * The first question to start the journey.
+ * @satisfies {Partial<ServerRoute>}
+ */
+export const onOffFarmGetController = {
+  handler(req, h) {
+    req.yar.set('animal', 'cow')
```
hughfdjackson, line 7: I think this was some test code that could be removed?

src/server/on-off-farm/controller.js
```diff
@@ -0,0 +1,48 @@
...
+    const onOffFarm = req.yar.get('onOffFarm')
```
hughfdjackson, line 8: Is this not asynchronous? If we're doing a round-trip to redis, then I'd want to be awaiting here.

src/server/on-off-farm/controller.js
```diff
@@ -0,0 +1,48 @@
...
+
+    return h.view('on-off-farm/index', {
+      pageTitle: 'Are you moving the cattle on or off your farm?',
+      heading: 'Are you moving the cattle on or off your farm?',
```
hughfdjackson, line 12: I don't think we're using these - looks like they're hardcoded later on?

And for what it's worth, I think it makes sense that they are hardcoded in the template.  Unless we extract all question-specific elements & have a fully generic 'radio.njk' template, then we may as well have all of the question specific text in the template itself?

That is, of course, except for dynamic elements like errors or currently selected values.

src/server/on-off-farm/controller.js
```diff
@@ -0,0 +1,48 @@
...
+export const onOffFarmPostController = {
+  handler(req, res) {
+    const { onOffFarm } = req.payload
+
+    if (!onOffFarm) {
+      return res.view('on-off-farm/index', {
+        pageTitle: 'Are you moving the cattle on or off your farm?',
+        heading: 'Are you moving the cattle on or off your farm?',
+        errorMessage: {
+          text: 'Select if you are moving cattle on or off your farm?'
```
hughfdjackson, line 29: I think we need to drop the question mark here - looking at the component guidance, all of the examples are framed as statements rather than questions: https://design-system.service.gov.uk/components/error-message/

src/server/on-off-farm/controller.js
```diff
@@ -0,0 +1,48 @@
...
+        }
+      })
+    }
+
+    req.yar.set('onOffFarm', onOffFarm)
```
hughfdjackson, line 34: Expecting an await here again, if there's a spill-over to redis state.   Otherwise we're in 'save-and-pray' territory - the user will think (on failure to get to redis) that they've succeeded & continue to the next page, but in reality it'll be missing from their page.

I'm sure there's a reason we're seeing a synchronous interface here, but it isn't sitting right with me!

src/server/on-off-farm/controller.js
```diff
@@ -0,0 +1,48 @@
...
+
+    return res.view('on-off-farm/index', {
+      pageTitle: 'Are you moving the cattle on or off your farm?',
+      heading: 'Are you moving the cattle on or off your farm?',
+      onOffFarm: {
+        value: onOffFarm
+      }
+    })
+  }
+}
```
hughfdjackson, line 44: Do we have a generic fallback on error that makes the journey recoverable for users?

src/server/on-off-farm/controller.test.js
```diff
@@ -0,0 +1,64 @@
+import { createServer } from '~/src/server/index.js'
+import { statusCodes } from '~/src/server/common/constants/status-codes.js'
+
+describe('#onOffFarmController', () => {
+  /** @type {Server} */
+  let server
+
+  beforeAll(async () => {
+    server = await createServer()
```
hughfdjackson, line 9: During test, are we using persistent state (e.g. redis running on docker)?  As the service expands, we may want to start mocking out parts of this, but I'm definitely okay for the moment going with that.

hughfdjackson (changes requested):

src/server/on-off-farm/index.js
hughfdjackson, line 14: There may be a reason for this, but this could be easily cut down to just a list of routes, and then in server.js we can call server.routes([...onOffFarmRoutes]).

This one's a comment rather than a request for change.

src/server/on-off-farm/controller.js
DrogoNevets: i thought the same - but apparently not looking at the source
I suspect what _actually_ happens is that it stores it in memory, then flushes to REDIS as a persistence type thing
so REDIS is infact a backup almost

src/server/on-off-farm/controller.js
DrogoNevets: the higher level template (layout.njk i think its called) is using them

src/server/on-off-farm/controller.js
DrogoNevets: see above comments

src/server/on-off-farm/controller.js
DrogoNevets: there is a generic 500 page yes

src/server/on-off-farm/controller.test.js
DrogoNevets: Catch up with you later on how the cache works - it will be using an in memory cache here i suspect

eoin-corr-git:
Change browser title from:
'Are you moving the cattle on or off your farm? | Applications and permissions' 

to 

'Are you moving the cattle on or off your farm?'

eoin-corr-git:
Error box stretches too far off page and missing question marks on end of sentence

eoin-corr-git (approved):

PR #9, opened 2024-11-13
DSFAAP-452: CSRF protection

hughfdjackson:
Using Crumb, which:

* sets a session cookie with crumb=<some value>
* injects {{crumb}} as a variable into out templates for us to use
* checks the 'crumb' payload field matches the offered cookie


src/server/common/test-helpers/csrf.js
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * @param {OutgoingHttpHeaders} headers
+ * @param {string} name
+ * @returns {string}
+ */
+const findCookie = (headers, name) => {
+  const setCookie = headers['set-cookie']
+
+  /** @type {string[] | undefined} */
+  const cookies = typeof setCookie === 'string' ? [setCookie] : setCookie
```
DrogoNevets, line 13: given you are wanting to put it into an array
would `Array.isArray` not be more appropriate?

src/server/on-off-farm/controller.test.js
```diff
@@ -15,25 +20,27 @@ describe('#onOffFarmController', () => {
   })
 
   test('Should provide expected response', async () => {
-    const { result, statusCode } = await server.inject({
+    const { payload, statusCode } = await server.inject({
```
DrogoNevets, line 23: this is the raw HTML not payload, so not sure payload is the best name

DrogoNevets (changes requested):
I assume we are going to want this on every page
so would it not be better to define it at a more global level rather than on a route by route level?

src/server/common/test-helpers/csrf.js
hughfdjackson: If it's not an array, it's either undefined or a string, so I wrapped it in an array if it .. isn't an array, then we'd end up with `{string[] | undefined[]}` or something similar - which just felt odd.

But it's ugly as is, so I'm going pushing an alternative that's more direct.

src/server/on-off-farm/controller.test.js
hughfdjackson: We can use 'result', but result has a wider type than we'd want: 

```
export interface ServerInjectResponse<Result = object> extends ShotResponseObject {
    /**
     * the raw handler response (e.g. when not a stream or a view) before it is serialized for transmission. If not available, the value is set to payload. Useful for inspection and reuse of the
     * internal objects returned (instead of parsing the response string).
     */
    result: Result | undefined;
    /**
     * the request object.
     */
    request: Request;
}
```

Whereas payload is guaranteed to be a string: 

```
export interface ResponseObject {
    /**
     * An object containing the raw request and response objects.
     */
    raw: {
        /**
         * The simulated request object.
         */
        req: InjectedRequest;

        /**
         * The simulated response object.
         */
        res: ServerResponse;
    };

    /**
     * An object containing the response headers.
     */
    headers: OutgoingHttpHeaders;

    /**
     * The HTTP status code.
     */
    statusCode: number;

    /**
     * The HTTP status message.
     */
    statusMessage: string;

    /**
     * The payload as a UTF-8 encoded string.
     */
    payload: string;

    /**
     * The raw payload as a Buffer.
     */
    rawPayload: Buffer;

    /**
     * An object containing the response trailers
     */
    trailers: NodeJS.Dict<string>;
}
```

I don't disagree that the naming could be better, but this is the value we're after IMO.  Does that make sense to you?

hughfdjackson:
> I assume we are going to want this on every page
> so would it not be better to define it at a more global level rather than on a route by route level?

IMO, the implementation needs to be global, and the tests should be a sampling.  A balanced approach would probably be: 

1. a couple of question pages
2. the declare & submit page (because this is the most important step in a users' transaction with us)

I wouldn't advocate putting it in every single route test, but I also don't think it makes sense to put it only in one indicative test.

DrogoNevets (approved):

PR #10, opened 2024-11-14
[DFSAAP-447] added CPH number page

DrogoNevets:


eoin-corr-git:
Our text box is smaller than the one on the prototype

src/server/cph-number/controller.js
```diff
@@ -0,0 +1,68 @@
+export const pageTitle =
+  'What is the County Parish Holding (CPH) number of your farm or premises where the animals are moving off?'
+
+/**
+ * The first question to start the journey.
```
hughfdjackson, line 5: Worth removing this comment, I think!

src/server/cph-number/controller.js
```diff
@@ -0,0 +1,68 @@
...
+      heading: pageTitle,
+      cphNumber: {
+        value: cphNumber
+      }
+    })
+  }
+}
+
+/**
+ * Respond to the first question.
```
hughfdjackson, line 23: Same comment here

src/server/cph-number/controller.js
```diff
@@ -0,0 +1,68 @@
...
+ * @satisfies {Partial<ServerRoute>}
+ * @param req
+ */
+export const postController = {
+  handler(req, res) {
+    const valid = /([0-9]{2})\/([0-9]{3})\/([0-9]{4})/g
```
hughfdjackson, line 29: Not something to do now, but something to think about for the future: 

We're likely to have to re-validate at least some data on the declaration screen - e.g. when users submit data that depends on answer to previous questions.

There's nothing stopping a user from updating a single question (and invalidating some previously submitted state - such as updating the 'earliest' and 'latest' dates, making the latest date before the earliest).

When they do that, we'll have to detect what they'e made invalid so that we can guide them through those pages *and* prevent them from declaring and submitting if they deep link straight back to the summary page. 

Also, we have access to `joi`, so we might find there's a more direct way of embedding error messages in there.

hughfdjackson (approved):

hughfdjackson:
@DrogoNevets  - two updates needed, I think, comparing this to https://design-system.service.gov.uk/patterns/validation/

1. `Error:` prefix to the title
2. we should re-populate the form data with the (wrong) form data that the user submitted so that they can edit rather than re-enter.

> Do not clear any form fields when validating users’ answers. Keep both passing and failing answers. This is to comply with WCAG 2.2 success criterion [3.3.7 Redundant entry](https://www.w3.org/WAI/WCAG22/Understanding/redundant-entry.html).

PR #15, opened 2024-11-18
[UNKNOWN] generic exit page

DrogoNevets:
- **exit page started**
- **trying to get tests working --no-verify**
- **trying to get tests working**
- **updated tests for branch**
- **change title size**

### DIFF
--- a/src/server/common/constants/status-codes.js
+++ b/src/server/common/constants/status-codes.js
@@ -8,5 +8,6 @@ export const statusCodes = {
   unauthorized: 401,
   forbidden: 403,
   notFound: 404,
-  imATeapot: 418
+  imATeapot: 418,
+  redirect: 302
 }
--- a/src/server/cph-number/controller.js
+++ b/src/server/cph-number/controller.js
@@ -29,7 +29,10 @@ export const postController = {
   handler(req, res) {
     const valid = /([0-9]{2})\/([0-9]{3})\/([0-9]{4})/g
     let errorMessage = null
-    const { cphNumber } = /** @type {CphNumberPayload} */ (req.payload)
+    let { cphNumber } = /** @type {CphNumberPayload} */ (req.payload)
+
+    // Remove whitespace from cphNumber
+    cphNumber = cphNumber ? cphNumber.replace(/\s+/g, '') : ''
 
     if (!cphNumber) {
       errorMessage = {
--- a/src/server/cph-number/controller.js
+++ b/src/server/cph-number/controller.js
@@ -30,7 +30,9 @@ export const getController = {
 export const postController = {
   handler(req, res) {
     const { cphNumber } = /** @type {CphNumberPayload} */ (req.payload)
-    const [isValid, message] = validator(cphNumber)
+    // Remove whitespace from cphNumber
+    const input = cphNumber ? cphNumber.replace(/\s+/g, '') : cphNumber
+    const [isValid, message] = validator(input)
 
     if (!isValid) {
       req.yar.clear('cphNumber')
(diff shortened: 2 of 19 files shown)


src/server/common/constants/status-codes.js
```diff
@@ -8,5 +8,6 @@ export const statusCodes = {
   unauthorized: 401,
   forbidden: 403,
   notFound: 404,
-  imATeapot: 418
+  imATeapot: 418,
+  redirect: 302
```
hughfdjackson, line 12: I _think_ it's intuitive that this is the temporary redirect (and that we call `301` `permanantRedirect` in the future).  Do you think there's any scope for confusion?

src/server/cph-number/controller.js
```diff
@@ -28,7 +28,10 @@ export const postController = {
   handler(req, res) {
     const valid = /([0-9]{2})\/([0-9]{3})\/([0-9]{4})/g
     let errorMessage = null
-    const { cphNumber } = /** @type {CphNumberPayload} */ (req.payload)
+    let { cphNumber } = /** @type {CphNumberPayload} */ (req.payload)
+
+    // Remove whitespace from cphNumber
+    cphNumber = cphNumber ? cphNumber.replace(/\s+/g, '') : ''
```
hughfdjackson, line 34: happy to pull it through as one this time, but would prefer a separate PR for this typically.

src/server/exit-page/index.njk
```diff
@@ -0,0 +1,35 @@
+{% extends 'layouts/page.njk' %}
+
+{% block content %}
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <div class="govuk-body" data-testid="app-page-body">
+        <div class="gem-c-title govuk-!-margin-top-0 govuk-!-margin-bottom-8">
```
hughfdjackson, line 8: In the prototype, we're missing this wrapper & I can't see it in the [GDS design system](https://design-system.service.gov.uk/styles/headings/).  

Worth removing?

hughfdjackson (changes requested):
Just one real change here - the styling on the header.

src/server/cph-number/controller.js
hughfdjackson: @eoin-corr-git  - are we looking to add a test for this case in the user journey tests?

src/server/common/constants/status-codes.js
DrogoNevets: my concern over 301 was that it redirects to different places depending on the input (`payload`) and thus I thought it was more a 302 than a 301

Also looking at [MDN 302](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302) = Found, not temp (which was news to me, but why i chose it over 300)

src/server/cph-number/controller.js
DrogoNevets: totally agree!!

src/server/exit-page/index.njk
DrogoNevets: hmmm woder where i got this from....will have a look

src/server/common/constants/status-codes.js
hughfdjackson: @DrogoNevets  - yup, 100% agreed that 302 is the right choice here.  It's just about the wording of the status code in the map.

DrogoNevets: oh i see, can change it to `Found` as per MDN?

src/server/exit-page/index.njk
DrogoNevets: now follow [this layout](https://design-system.service.gov.uk/styles/layout/)

hughfdjackson (approved):

eoin-corr-git:
Page title should be 'This service is not available for your movement type'

PR #27, opened 2024-11-21
DSFAAP-516: make crumb cookie not secure for local development

joseluisgraa:


src/server/common/helpers/csrf-plugin.js
```diff
@@ -1,6 +1,11 @@
 import crumb from '@hapi/crumb'
+import { config } from '~/src/config/config.js'
 
 export const csrfPlugin = {
   plugin: crumb,
-  options: {}
+  options: {
+    cookieOptions: {
+      isSecure: config.get('session.cookie.secure')
```
DrogoNevets, line 8: do we need to think about the other cookie options here too? like TTL, httpOnly etc?

joseluisgraa: secure seems to be the only one causing the issue with safari and localhost so I suppose it is fine as it is

hughfdjackson: @DrogoNevets  - I don't think we need those to be configurable, so long as they default to the right values (e.g. definitely httpOnly, or the protection this gives us disappears)

joseluisgraa: Seems to be using the default hapi values, whatever they are. Checking what those are

joseluisgraa: - **ttl** (time to live): null, which means the cookie will be deleted when the browser is closed.
- **isSecure**: true, which means the cookie will only be transmitted over a secure protocol (HTTPS).
- **isHttpOnly**: true, which means the cookie will be inaccessible to JavaScript and can only be accessed by the web server.
- **encoding**: 'none', which means the cookie value will not be encoded.
- **clearInvalid**: false, which means invalid cookies will not be automatically cleared.
- **strictHeader**: true, which means the cookie will be set with the Strict attribute, indicating that it should only be sent over a secure protocol.

joseluisgraa: Shall I set TTL to config.get('session.cookie.ttl')?

hughfdjackson: For this change - so we don't stand in the way of progress - i think we should go for this PR as is.  We can raise a separate PR if later we think / determine the TTL needs to be updated.

PR #28, opened 2024-11-21
BAU: extract validator as a standardised function

hughfdjackson:
We're going to have to revalidate pages in a number of contexts:

* on initial submission
* when on the 'check answers' page
* when on the task list (e.g. a task is only complete if it has a complete suite of valid answers)
* when on final declare & submit page

Starting to move towards a uniform interface that can validate a form submission or state is a first step towards making that easier.

### DIFF
--- a/src/server/common/page/model.js
+++ b/src/server/common/page/model.js
@@ -1,0 +1,10 @@
+/**
+ * @template Payload
+ * @template Data
+ * @typedef {{
+ *   clean: (payload: Payload) => Data,
+ *   validate: (data: Data) => ValidationResult
+ * }} Model<Payload, Data>
+ */
+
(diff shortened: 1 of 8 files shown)


src/server/common/page/model.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * @template Payload
+ * @template Data
+ * @typedef {{
+ *   clean: (payload: Payload) => Data,
+ *   validate: (data: Data) => ValidationResult
+ * }} Model<Payload, Data>
+ */
+
+/** @import {ValidationResult} from './validation.js' */
```
hughfdjackson, line 10: I didn't mean to add this - but this is give or take where I see the direction of travel.

PR #32, opened 2024-11-25
[DSFAA-455] origin summary page and other question redirects

DrogoNevets:
- **lint fix**
- **added simplified session helper**
- **added negative tests**
- **added tests**


src/server/origin/summary/controller.js
```diff
@@ -0,0 +1,45 @@
...
+    const originOnOffFarm = origin?.onOffFarm
+    let enteredOnOffFarm = ''
+
+    if (originOnOffFarm === 'on') {
+      enteredOnOffFarm = 'On to the farm or premises'
+    } else if (originOnOffFarm === 'off') {
+      enteredOnOffFarm = 'Off the farm or premises'
+    } else {
+      enteredOnOffFarm = ''
+    }
```
hughfdjackson, line 23: At some stage, I'm keen to move this into the session state so it can be looked up dynamically.

we'll be re-using this in the overall summary, and then again when we send that data for review by the admin staff, so it's important to have a common place to access it from.

I think this is probably a target for a later refactoring though?

src/server/origin/summary/index.njk
```diff
@@ -0,0 +1,67 @@
...
+
+{% block questions %}
+  <h1 class="govuk-heading-l">{{pageTitle}}</h1>
+
+  {{ govukSummaryList({
+    classes: "govuk-!-margin-bottom-9",
+    rows: [
+      {
+        key: {
+          text: "Are you moving the cattle on or off your farm or premises?"
```
hughfdjackson, line 16: Similar to the above, this feels ripe for consolidation into something you pull from the session state. 

The reason I'm saying state instead of component or model is because later questions will be dynamic - e.g.: in R2 or R3 we will be asking about dilution ratios of the specific disinfectant you're using.  At that stage, the question text isn't static for a given page, it's generated based on prior state.

As with the others, I'm happy to merge for functionality & then refactor to address.

src/server/common/helpers/next-page.js
```diff
@@ -0,0 +1,20 @@
...
+    const { pathname } = new URL(referrerURI)
+    return {
+      nextPage: pathname === needle ? '/origin/summary' : ''
+    }
+  } catch (e) {
+    return {
+      nextPage: null
+    }
+  }
+}
```
hughfdjackson, line 16: This approach feels fragile to me, as discussed on slack!


> the common alternative is using redirect_uri query param - so the page state depends on session state & url
> in the referrer version, it's session state, url & the referrer (which doesn't survive copy & pasting for example)

This is the one thing I think we should change, unless there're pros to the referrer version that I'm missing.

hughfdjackson (changes requested):

src/server/origin/address/controller.js
```diff
@@ -11,12 +11,12 @@ export const pageHeading =
...
   handler(req, h) {
-    const originAddress = req.yar.get('originAddress')
+    const { address } = req.yar.get('origin') ?? {}
 
     return h.view(indexView, {
       pageTitle,
       heading: pageHeading,
-      values: originAddress
+      values: address
     })
```
hughfdjackson, line 20: I'm expecting to see a `nextPage` for this, pulled from the redirect_uri?

Worth putting in a corresponding test too, IMO.

src/server/origin/address/controller.js
```diff
@@ -50,6 +50,7 @@ export const originAddressPostController = {
     }))
 
     if (!isValid) {
+      req.yar.clear('originAddress')
```
hughfdjackson, line 53: `originAddress` is no longer the session field - it's `origin.address` isn'tit?

hughfdjackson: I'm not 100% sure we should clear state on failure, FWIW. 

Imagine the scenario where: 

1. user is on the summary page
2. user clicks `change` against address
3. user attempts a change and gets an error back
4. user navigates back to summary page

I'm not sure if it's most intuitive that the address is wiped out, or if it's what it was before a failed attempt.

src/server/origin/address/controller.test.js
```diff
@@ -257,19 +259,25 @@ describe('#originAddressController', () => {
...
+    url: '/origin/address'
   })
 
   testCsrfProtectedPost(() => server, {
     method: 'POST',
-    url: '/origin-address',
+    url: '/origin/address',
     payload: {
-      originAddress: 'on'
+      originAddress: {
```
hughfdjackson, line 274: I know it isn't relevant here, but the payload for POST will always be an object with: 

```
{ [key: string]: string }
```

because that's the form data payload that the framework's expecting.

src/server/origin/cph-number/controller.test.js
```diff
@@ -84,6 +84,37 @@ describe('#cphNumber', () => {
...
+        payload: {
+          cphNumber: '12/123/1234',
+          nextPage: '/origin/summary'
+        }
+      })
+    )
+
+    expect(headers.location).toBe('/origin/summary')
+    expect(statusCode).toBe(statusCodes.redirect)
+  })
```
hughfdjackson, line 117: nice :D

src/server/origin/cph-number/controller.js
```diff
@@ -48,13 +51,17 @@ export const postController = {
...
     }
 
-    req.yar.set('cphNumber', input)
+    req.yar.set('origin', {
+      ...req.yar.get('origin'),
+      cphNumber: input
+    })
 
-    return res.redirect('/origin-address')
+    return res.redirect(nextPage ?? '/origin/address')
```
hughfdjackson, line 59: I think this is broken, because in all cases the nextPage hidden field is going to be "" if a redirect_uri is not provided.

so the null coalescing causes this to return a 302 with location of `''`

src/server/origin/on-off-farm/controller.js
```diff
@@ -39,11 +42,14 @@ export const onOffFarmPostController = {
...
-    req.yar.set('onOffFarm', onOffFarm)
+    req.yar.set('origin', {
+      ...req.yar.get('origin'),
+      onOffFarm
+    })
 
     switch (onOffFarm) {
       case 'off':
-        return res.redirect('/cph-number')
+        return res.redirect(nextPage ?? '/origin/cph-number')
```
hughfdjackson, line 52: Same bug here as above - nextPage isn't going to be null / undefined in any circumstance, so we're getting a redirect to a blank location.

src/server/origin/summary/controller.test.js
```diff
@@ -0,0 +1,177 @@
...
+        expect.stringContaining('Off the farm or premises')
+      )
+
+      expect(payload).toEqual(expect.stringContaining('12/123/1234'))
+      expect(payload).toEqual(
+        expect.stringContaining(
+          Object.values(defaultState.address).join('<br />')
+        )
+      )
+    })
```
hughfdjackson, line 104: Since all fields are mandatory in this section, I'm expecting to see a redirect to the `/origin/to-or-from-own-premises` page here, as per our slightly vague AC: 


```
Scenario5: User find check your answer page before any answers have been provided 
When this happens, do as defra forms do
```

hughfdjackson: Well, technically that's before `any`, but I believe the spirit we're getting at is "if you haven't met the pre-requisites when you're on this page, you're taken to all of the relevant questions until you return to this page in a good state"

hughfdjackson (changes requested):

src/server/origin/address/controller.js
DrogoNevets: didnt put it here as this will also go to the summary page as is the last page in the flow - so whther it comes from summary or not, it will also go there, didnt see the point of the added complexity for no reason

src/server/origin/address/controller.js
DrogoNevets: i would argue thats less intuitive, theres a chance user might think they have changed the address or at least said "do not go to the address previously entered" anymore, but if we keep the previously entered one then we have no way of checking the address is in fact what is intended and mark as valid

src/server/origin/address/controller.test.js
DrogoNevets: this locks it down to `{{
 *  addressLine1: string;
 *  addressLine2 ?: string;
 *  addressTown: string;
 *  addressCounty ?: string;
 *  addressPostcode: string;
 * }}` though meaning that we know what keys we are fetching - if we accept the "any key string" principle, then there is no need to have typescript as it could be anything, we are saying its not just anything

src/server/origin/address/controller.js
hughfdjackson: But also, if we're going down this path then I suspect we should add it to all cases?

src/server/origin/address/controller.js
hughfdjackson: I can see the logic there - I suspect this'll get factored out a bit later.

src/server/origin/address/controller.js
DrogoNevets: have just done so in the commits below too (i noticed the same thing)

src/server/origin/address/controller.test.js
hughfdjackson: Sorry, this was poorly phrased. 

the payload isn't nested within `originAddress` - it's just fields beneath that.  e.g.

```javascript
{ 
    addressLine1: 'Starfleet Headquarters',
    addressLine2: '24-593 Federation Drive',
    addressTown: 'San Francisco',
    addressCounty: 'San Francisco',
    addressPostcode: 'RG24 8RR'
}
```
rather than what it currently has:

```javascript
{ 
  originAddress: { 
    addressLine1: 'Starfleet Headquarters',
    addressLine2: '24-593 Federation Drive',
    addressTown: 'San Francisco',
    addressCounty: 'San Francisco',
    addressPostcode: 'RG24 8RR'
  }
}
```

(forgive my crimes against indentation, the comment editor isn't great at that.

hughfdjackson (approved):

src/server/origin/on-off-farm/validator.js
```diff
@@ -0,0 +1,30 @@
...
+      'Select if you are moving cattle on or off your farm or premises',
+    'any.valid':
+      'Select if you are moving cattle on or off your farm or premises'
+  })
+
+/**
+ * @param {string} cphNumber
+ * @returns {{isValid: boolean, errors: object}}
+ */
+export default (cphNumber) => {
```
joseluisgraa, line 17: This should be onOffFarm I guess

src/server/origin/on-off-farm/validator.js
```diff
@@ -0,0 +1,30 @@
...
+  const result = CphNumberSchema.options({ abortEarly: false }).validate(
+    cphNumber
+  )
+  const errors = result.error?.details.map(({ message }) => [
+    'cphNumber',
```
joseluisgraa, line 22: onOffFarm

src/server/origin/cph-number/controller.js
```diff
@@ -29,32 +31,41 @@ export const getController = {
...
 
     if (!isValid) {
-      req.yar.clear('cphNumber')
+      req.yar.set('origin', {
+        ...req.yar.get('origin'),
+        cphNumber: undefined
+      })
+
       return res.view(indexView, {
+        nextPage,
```
joseluisgraa, line 48: I've added nextPage here as otherwise after an error the redirect was not happening. I've also added a unit test for it. A similar change is needed in the other controllers after an error.

PR #39, opened 2024-11-28
DSFAAP-510: create task list page and add model for tasks

joseluisgraa:


hughfdjackson:
@joseluisgraa  - I think there's something off with the structure here. 

The models (e.g. `Address`) represent datatypes that could appear in multiple sections.  For instance, we collect an `Address` in both the "Movement Origin" section & the "Receiving the Licence" section.  

We want the user experience (in terms of validation, error messages, etc) to be consistent across those sections - which the `Address` model allows for, if it's reused. 

A better layout would be (IMO): 

`~/src/server/common/answer/` (containing the existing models)

and

`~/src/server/common/section/` (containing the new section models)

Then the origin & licensed sections can make reference to the Address answer model.

src/server/common/model/destination/destination.js
```diff
@@ -0,0 +1,36 @@
+import Joi from 'joi'
+import { Model, validateAgainstSchema } from '../model.js'
+
+const validationSchema = Joi.object().required()
+
+/**
+ * @import {RawPayload} from '../model.js'
+ */
+
+export class Destination extends Model {
```
hughfdjackson, line 10: I'd be tempted to make a SectionModel that only has the interface we need for sections. 

At this point in time I think it just needs `fromState` and `validate`?

src/server/common/model/origin/cph-number.js
```diff
@@ -45,8 +45,12 @@ export class CphNumber extends Model {
...
-    return new CphNumber({
-      cphNumber: state
-    })
+    if (state) {
+      return new CphNumber({
+        cphNumber: state
+      })
+    } else {
+      return new CphNumber({})
+    }
```
hughfdjackson, line 54: good catch!

hughfdjackson: (worth test coverage too)

joseluisgraa:
You are right, I created the one for the task and then started moving the existing ones into origin without realising they would be reused by different sections. I'll follow your proposed structure.

src/server/task-list-incomplete/controller.js
```diff
@@ -0,0 +1,23 @@
+const pageTitle =
+  'You need to complete all of the sections before you review and submit'
+const heading =
+  'You need to complete all of the sections before you review and submit'
```
hughfdjackson, line 4: `const heading = pageTitle`?

hughfdjackson (changes requested):

src/server/common/model/answer/answer-model.test.js
```diff
@@ -0,0 +1,69 @@
+import Joi from 'joi'
+import { AnswerModel, validateAgainstSchema } from './answer-model.js'
+
+describe('validateAgainstSchema', () => {
+  let validationSchema
+
+  beforeEach(() => {
+    validationSchema = Joi.object({
```
hughfdjackson, line 8: schemas don't get mutated by being validated (may even be immutable?), so you can do: 

```javascript
const validationSchema = Joi.object({
  addressLine1: Joi.string().required(),
  postcode: Joi.string().required()
})
```

and drop the `beforeEach` block

eoin-corr-git (approved):

PR #40, opened 2024-11-28
run-bs-tests

eoin-corr-git:


.github/workflows/run-tests-browserstack.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  id-token: write
+  contents: write
+  pull-requests: write
```
hughfdjackson, line 11: Since we're running this on `main`, presumably we don't need this permission?

.github/workflows/run-tests-browserstack.yaml
```diff
@@ -0,0 +1,38 @@
+name: Browserstack tests
+
+on:
+  push:
+    branches:
+      - main
+
+permissions:
+  id-token: write
+  contents: write
```
hughfdjackson, line 10: Just looking at [the permissions](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token), I'm not sure we'll need this? 

```
Work with the contents of the repository. For example, contents: read permits an action to list the commits, and contents: write allows the action to create a release. For more information, see "[Permissions required for GitHub Apps](https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-contents)."
```

Possibly `contents:read`

hughfdjackson (changes requested):

hughfdjackson (approved):

PR #47, opened 2024-12-03
[514] check your answers

DrogoNevets:


src/server/check-answers/controller.test.js
```diff
@@ -0,0 +1,143 @@
+import { createServer } from '~/src/server/index.js'
+import { statusCodes } from '~/src/server/common/constants/status-codes.js'
+import { pageTitle } from './controller.js'
+import { withCsrfProtection } from '~/src/server/common/test-helpers/csrf.js'
+import { parseDocument } from '~/src/server/common/test-helpers/dom.js'
+import SesessionTestHelper from '../common/test-helpers/session-helper.js'
```
hughfdjackson, line 6: `SessionTestHelper`?

src/server/check-answers/controller.test.js
```diff
@@ -0,0 +1,143 @@
...
+    expect(payload).toEqual(
+      expect.stringContaining(
+        '<a href="#confirmation">You need to tick a declaration box</a>'
+      )
+    )
+
+    expect(statusCode).toBe(statusCodes.ok)
+  })
+
+  it('Should redirect to itself when there is no error', async () => {
```
hughfdjackson, line 95: convention we've been talking about recently is redirecting on to the destination page, and letting that page 404 (until the next ticket).

src/server/check-answers/controller.test.js
DrogoNevets: yes - this is the helper i built a couple of weeks ago to store a session within the test so we can use it

src/server/check-answers/index.njk
```diff
@@ -0,0 +1,129 @@
...
+          {% endif %}
+          <h1 class="govuk-heading-l">
+              {{ heading }}
+          </h1>
+
+          <h2 class="govuk-heading-m">Movement origin</h2>
+          {{ section([{
+            key: "Are you moving the cattle on or off your farm or premises?", 
+            value: "", 
+            url: "/origin/summary"
```
hughfdjackson, line 65: The urls here should take you to the question you want to change directly, rather than the summary page.

DrogoNevets: and the summary page should take you to the first question it finds that is invalid

DrogoNevets: thus meaning we dont have to duplicate logic

hughfdjackson: This isn't about invalid pages - you could have valid data you wish to change.  e.g. your CPH may be valid, but you may realise it's the wrong one and click on 'change'

joseluisgraa (approved):

PR #48, opened 2024-12-03
Fix iphone capabilities

eoin-corr-git:


user-journey-tests/specs/origin/checkAnswers.spec.js
```diff
@@ -33,7 +33,7 @@ describe('Check your answers test', () => {
     await loadPageAndVerifyTitle('', landingPage.pageTitle)
   })
 
-  it('Should verify the back link is history -1', async () => {
+  it.only('Should verify the back link is history -1', async () => {
```
joseluisgraa, line 36: you forgot to remove this .only here?

user-journey-tests/page-objects/page.js
```diff
@@ -84,7 +84,11 @@ class Page {
   }
 
   async selectContinue() {
+    const deviceName = browser.capabilities?.deviceName
     await page.selectElement(this.getContinueButton())
+    if (deviceName === 'iPhone 13') {
```
joseluisgraa, line 89: Could define this as a constant somewhere, or even as an array of device names that need the check for security popup (only iphone 13 for now) and you check if the array includes deviceName

user-journey-tests/helpers/page.js
```diff
@@ -91,3 +91,15 @@ export const clearElement = async (element) => {
     )
   }
 }
+
+export const checkForSecurityPopUpAndResolve = async () => {
+  const contexts = await browser.getContexts()
+  if (contexts.length > 1) {
```
joseluisgraa, line 97: As this is checking whether we have more than one context, I wonder if you can remove the check for device name and run them all past this function?

eoin-corr-git: It only recognises this contexts function in browserstack so otherwise it errors the test because it doesn't recommend the command

eoin-corr-git: browser.getContexts is not a function

joseluisgraa (approved):

PR #49, opened 2024-12-03
DSFAAP-520: add check answers page for license flow

joseluisgraa:


src/server/common/model/section/license.test.js
```diff
@@ -9,16 +12,45 @@ describe('License', () => {
...
+
+    it('should return invalid if any nested object is invalid', () => {
+      const licenseData = {
+        emailAddress: ''
+      }
+
+      const result = License.fromState(licenseData).validate()
+
       expect(result.isValid).toBe(false)
+      expect(result.errors).toHaveProperty('license')
```
hughfdjackson, line 33: ultimately, we're going to want this to be the key of the question that's erroring: 

```javascript
{
  emailAddress: { text: "...." }
}
```

In more complex sections, that will give us the data we need to know which question route users to.

src/server/license/check-answers/controller.js
```diff
@@ -0,0 +1,45 @@
...
+/**
+ * @satisfies {Partial<ServerRoute>}
+ */
+export const licenseCheckAnswersGetController = {
+  handler(req, h) {
+    const license = req.yar.get('license') ?? {}
+
+    const emailAddress = EmailAddress.fromState(
+      license?.emailAddress
+    ).validate()
```
hughfdjackson, line 17: If we make the change as suggested above, this is a great place to use LicenceSection.fromState(req.yar.get('licence')

src/server/license/check-answers/index.njk
```diff
@@ -0,0 +1,34 @@
...
+
+  {{ govukSummaryList({
+    classes: "govuk-!-margin-bottom-9",
+    rows: [
+      {
+        key: {
+          text: "What email address would you like the licence sent to?"
+        },
+        value: {
+          text: license.emailAddress
```
hughfdjackson, line 17: this should be the html() method of the `emailAddress` - the one that's designed for rendering a summary, not necessarily for playing back into an input field.

hughfdjackson (changes requested):

src/server/common/model/section/license.test.js
joseluisgraa: Done

src/server/license/check-answers/controller.js
joseluisgraa: Done

hughfdjackson (approved):

PR #51, opened 2024-12-05
DSFAAP-577: filter payload in answer model

hughfdjackson:
To ensure only the payload fields we care about are captured in the state, we need to filter them down. 

Otherwise we'll capture things like nextPage & potentially show them to users.

### DIFF
--- a/src/server/common/model/answer/address.js
+++ b/src/server/common/model/answer/address.js
@@ -98,7 +98,7 @@ export class Address extends AnswerModel {
     return validateAnswerAgainstSchema(addressPayloadSchema, this._data ?? {})
   }
 
-  extractFields({
+  _extractFields({
     addressLine1,
     addressLine2,
     addressTown,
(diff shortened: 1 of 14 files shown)


src/server/common/model/answer/address.js
```diff
@@ -79,6 +79,7 @@ export class Address extends AnswerModel {
 
   get html() {
     return Object.values(this.value ?? [])
+      .filter((line) => line !== undefined)
```
hughfdjackson, line 82: The process of unpacking & repacking an object causes: 

```
{ addressLine1: 'fooBar', addressTown: 'London', addressPostcode: 'N11NB' }
```

into: 

```
{ addressLine1: 'fooBar', addressLine2: undefined, addressTown: 'London', addressCounty: undefined, addressPostcode: 'N11NB' }
```

I don't think this matters for us, other than in this context - but happy to be challenged.

NB - the reason I didn't use lodash's `pick` is because it isn't typesafe.

joseluisgraa (approved):

PR #54, opened 2024-12-06
Initial version for application model

joseluisgraa:


src/server/check-answers/controller.js
```diff
@@ -21,9 +22,13 @@ export const checkAnswersGetController = {
       license: License.fromState(req.yar.get('license'))
     }
 
-    const isValid = Object.values(tasks).every(
-      (task) => task.validate().isValid
-    )
+    // TODO see how to get the whole state in one go with req.yar.get('application')
```
joseluisgraa, line 25: Looked at the state in redis and it is under an "item" property but couldn't get that one so maybe we have to nest everything under another property?

src/server/common/model/application/application.test.js
```diff
@@ -0,0 +1,43 @@
...
+describe('Application', () => {
+  it('should create an Application instance from a valid state', () => {
+    const state = {
+      origin: originDefaultState,
+      license: licenseDefaultState
+    }
+
+    const application = Application.fromState(state)
+
+    expect(application).toBeInstanceOf(Application)
```
joseluisgraa, line 29: I would need to review these tests as it doesn't look great that I have to check _data.origin._data...

src/server/common/model/application/validation.js
```diff
@@ -0,0 +1,33 @@
...
+ *   result: ValidationResult
+ * }} ApplicationValidationResult
+ */
+
+/**
+ * Validates all the answers within the application and outputs an overall isValid result.
+ * @param {{[key:string]: SectionModel}} data - The data to be validated.
+ * @returns {ApplicationValidationResult} An object containing the validation result
+ */
+export const validateApplication = (data) => {
```
joseluisgraa, line 19: This validation function is identical to the one we have for section but not sure where we should be moving it to

src/server/check-answers/controller.js
hughfdjackson: @joseluisgraa  - I didn't want to pile in too much change at once, but I definitely thought that on `application` state that had the sections beneathe it ultimately makes sense. 

One for after this round of refactors & a bit of feature implementation?

src/server/common/model/application/application.test.js
hughfdjackson: agreed that this is breaking encapsulation a little too much - but I definitely don't think that's a blocker for now.

src/server/common/model/application/validation.js
hughfdjackson: BTW, I put in a refactor for the section version of this - mind to use that version for now?

hughfdjackson (changes requested):

src/server/check-answers/controller.js
hughfdjackson: yup, exactly that :)

hughfdjackson (approved):

PR #56, opened 2024-12-09
[REFACTOR] on-off-farm uses new models

DrogoNevets:


src/server/common/controller/question-page-controller.test.js
```diff
@@ -247,6 +247,30 @@ describe('QuestionPageController', () => {
     expect(statusCode).toBe(statusCodes.ok)
   })
 
+  test('Should display an error and set next page appropriately', async () => {
```
joseluisgraa, line 250: use "it" to keep it consistent

src/server/common/helpers/not-implmented-error.js
```diff
@@ -0,0 +1,5 @@
+export class NotImplementedError extends Error {
```
joseluisgraa, line 1: Can we rename the file as it has a typo?

src/server/common/model/answer/on-off-farm.js
```diff
@@ -45,6 +47,16 @@ export class OnOffFarm extends AnswerModel {
     }
   }
 
+  /** @param {AnswerErrors} errors */
+  static errorMessages(errors) {
```
joseluisgraa, line 51: Are we not supposed to use the static method defined already in the answer model?

src/server/common/model/answer/on-off-farm.test.js
```diff
@@ -73,6 +73,17 @@ describe('#OnOffFarm.fromState', () => {
   })
 })
 
+describe('OnOffFarm.errorMessages', () => {
```
joseluisgraa, line 76: If we just use the static method from answer model we might want to also unit test it in the model only?

src/server/origin/on-off-farm/index.js
```diff
@@ -1,36 +1,52 @@
-import {
-  onOffFarmGetController,
-  onOffFarmPostController
-} from './controller.js'
-
 /**
- * Sets up the routes used in the home page.
+ * Sets up the routes used in the cph number page.
```
joseluisgraa, line 2: on-off-farm not cph number

src/server/common/model/answer/on-off-farm.js
DrogoNevets: no as validation might be slightly different on each answer (one field vs many) so base throws a "not Implemented" error

src/server/common/model/answer/on-off-farm.test.js
DrogoNevets: see above

src/server/origin/on-off-farm/index.js
```diff
@@ -1,36 +1,52 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  nextPage(answer) {
+    if (answer.value === 'on') {
+      return exitPage
     }
+    return cphNumberPage
+  }
+
+  /** @param {AnswerErrors} errors */
+  errorMessages(errors) {
```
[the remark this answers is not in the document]
DrogoNevets: as above, its different for multi field vs single

PR #57, opened 2024-12-09
Separate bstack run

eoin-corr-git:


user-journey-tests/wdio.browserstack.extended.conf.js
```diff
@@ -0,0 +1,102 @@
...
+          projectName: 'apha-apps-perms-move-animal-ui',
+          buildName: 'apha-test-run'
+        },
+        acceptInsecureCerts: true,
+        forceLocal: true,
+        browserstackLocal: true
+      }
+    ]
+  ]
+})
```
hughfdjackson, line 102: To keep things as easy to manage as possible, is it worth replacing only the capabilities array?

```javascript
import { config as browserstackBaseConfig } from './wdio.browserstack.conf.js'

export const config = {...browserstackBaseConfig, 
  capabilities: [ ... ]
}
```

eoin-corr-git: It still needs the primary wdio.conf object as the file you mentioned doesn't include everything. Plus I think this approach gives testing the workflows more flexibility

hughfdjackson: 'wdio.browserstack.conf.js' already includes all the base config, since it does its own merge?

hughfdjackson: On the flexibility point - we can always move towards more divergence later if necessary.  But while they are intentionally similar, this tells us that without having to manually diff two configuration files that are meant to be largely similar IMO.

hughfdjackson (approved):

PR #63, opened 2024-12-10
[583] construct summary page dynamically

DrogoNevets:


src/server/common/model/answer/answer-model.js
```diff
@@ -67,6 +68,11 @@ export class AnswerModel {
     throw new NotImplementedError()
   }
 
+  /** @returns {Page} */
+  nextPage() {
+    throw new NotImplementedError()
+  }
```
hughfdjackson, line 74: The answer models are intended to be the more re-usable part - e.g. Address, Email address or name being generic - so the nextPage is on the Page abstraction as a result.

src/server/origin/summary/controller.js
```diff
@@ -13,6 +14,15 @@ export const originSummaryGetController = {
     const origin = Origin.fromState(req.yar.get('origin'))
     const { isValid, result } = origin.validate()
 
+    const pages = []
+
+    /** @type any */
```
hughfdjackson, line 19: Without the squigglies, jsdoc ignores this - fallen foul of this many times!

Also, could the type here be Page?  So 

```/** @type Page */```

File: src/server/origin/summary/controller.js [hunk 3]
```diff
@@ -27,14 +37,32 @@ export const originSummaryGetController = {
...
+            href: `${page.urlPath}?redirect_uri=/origin/summary`,
+            text: 'Change',
+            visuallyHiddenText: page.question,
+            attributes: {
+              'data-testid': `${page.questionKey}-change-link`
+            }
+          }
+        ]
+      }
+    }))
```
hughfdjackson, line 60: This is looking good!

hughfdjackson (changes requested):
Change requested is mainly about keeping Answers fully agnostic of their page context.

V. minor one about types

hughfdjackson (approved):

PR #65, opened 2024-12-11
[583-2] check question uses the models

DrogoNevets:
- **task list uses models**
- **removed uneeded**
- **view logic moved to view**

### DIFF
--- a/src/server/common/model/section/section-model.js
+++ b/src/server/common/model/section/section-model.js
@@ -20,14 +20,16 @@ export class SectionModel {
   /** @type {string}  */
   summaryPageLink
 
-  /** @type {boolean}  */
-  isEnabled = true
-
   /** @type {SectionPayload} */
   _data
 
   constructor(data) {
     this._data = data
+  }
+
+  /** @returns {boolean}  */
+  get isEnabled() {
+    throw new NotImplementedError()
   }
 
   seal() {
--- a/src/server/common/model/section/section-model.js
+++ b/src/server/common/model/section/section-model.js
@@ -3,6 +3,7 @@ import { validateSection } from './validation.js'
 
 /**
  * @import {AnswerModel} from '../answer/answer-model.js'
+ * @import { QuestionPage } from '../page/question-page-model.js'
  */
 
 /**
@@ -10,11 +11,26 @@ import { validateSection } from './validation.js'
  */
 
 export class SectionModel {
+  /** @type {string}  */
+  title
+
+  /** @type {QuestionPage}  */
+  initialPage
+
+  /** @type {string}  */
+  summaryPageLink
+
+  /** @type {boolean}  */
+  isEnabled = true
+
   /** @type {SectionPayload} */
   _data
 
   constructor(data) {
     this._data = data
+  }
+
+  seal() {
     Object.seal(this)
   }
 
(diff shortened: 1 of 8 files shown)


src/server/task-list/controller.js
```diff
@@ -13,63 +13,29 @@ const buttonText = 'Review and submit'
...
-      title: 'Movement destination',
-      initialLink: '#',
-      summaryLink: '#',
-      isValid: destination.validate().isValid,
-      isEnabled: isOriginValid
-    })
+    const application = {
+      origin: Origin.fromState(req.yar.get('origin')),
+      destination: Destination.fromState(),
+      tests: Tests.fromState(),
```
hughfdjackson, line 19: I believe we've dropped the conditional values that @joseluisgraa  has implemented.

For instance, the "Movement destination" section (aka `destination`) is enabled if the "Movement Origin" section (aka `origin`) is valid. 

Same dependency on `origin` from `tests`.

src/server/task-list/controller.js
```diff
@@ -13,63 +13,29 @@ const buttonText = 'Review and submit'
...
+    const tasks = Object.values(application)
 
-    const licenceGdsTask = buildGdsTaskItem({
-      title: 'Receiving the licence',
-      initialLink: '/receiving-the-licence/licence-enter-email-address',
-      summaryLink: '/receiving-the-licence/check-answers',
-      isValid: licence.validate().isValid,
-      isEnabled: true
+    const gdsTasks = tasks.map((section) => {
+      return {
```
hughfdjackson, line 26: should this be

```javascript
return buildGdsTaskItem({
  ...
})
```

The template expects something with a substantially different shape than we're passing in here.

src/server/task-list/index.njk
```diff
@@ -3,6 +3,60 @@
...
+    })) %}
+  {% endfor %}
+
+
+
+  {{ govukTaskList({
+    idPrefix: "task-list",
+    items: summaryRows
+  }) }}
+{% endmacro %}
```
hughfdjackson, line 58: TBH, I think readability of this version has taken a hit - vs the readability of the javascript version it replaces.

src/server/common/model/section/section-model.js
```diff
@@ -3,18 +3,34 @@ import { validateSection } from './validation.js'
...
+  title
+
+  /** @type {QuestionPage}  */
+  initialPage
+
+  /** @type {string}  */
+  summaryPageLink
+
+  /** @type {boolean}  */
+  isEnabled = true
```
hughfdjackson, line 24: In the current pattern, we're using these fields for configuration that doesn't change amongst object instances - e.g. all instances of this class will have this initialPage & summaryPageLink. 

In this case I don't think that's the case, since we're dynamically enabling this section in some cases.

src/server/task-list/controller.js
DrogoNevets: that is to be added to section page model
isEnabled could just as easily be a function that works out if its enabled or not

at present, we dont have a section model that is able to look at a section outside of itself

src/server/task-list/index.njk
DrogoNevets: whilst is half see your point, i dont think the controller should know about css classes etc - that is entirely view logic and has no place anywhere but the view

src/server/common/model/section/section-model.js
DrogoNevets: as above, more than happy for this to be a getter, that can dynamically fetch if its enabled or not

PR #72, opened 2024-12-13
DSFAAP-535: exit page solution sketch

hughfdjackson:
Basic approach is: 

- discover if the final page is an ExitPage, and if so: 
  - set the section as invalid (making it show incomplete on the task list, preventing 'Review & Submit', preventing the task summary from showing)
  - redirect the user back to the question *just before* the exit page, so the user can make a different choice if they choose
  
So that the user isn't sent back to the summary after they've chosen a condition that would cause them to exit, an exit page routing takes precedence over a redirect_uri.


src/server/common/controller/question-page-controller.js
```diff
@@ -77,8 +78,13 @@ export class QuestionPageController {
       ...req.yar.get(this.page.sectionKey),
       [this.page.questionKey]: answer.toState()
     })
-    return h.redirect(
-      calculateNextPage(payload.nextPage, this.page.nextPage(answer).urlPath)
-    )
+
+    const nextPage = this.page.nextPage(answer)
+
+    if (nextPage instanceof ExitPage) {
```
joseluisgraa, line 84: Should this not be also part of the calculateNextPage function?

src/server/common/controller/question-page-controller.test.js
```diff
@@ -149,6 +158,30 @@ describe('QuestionPageController', () => {
       expect(state.someOtherQuestion).toBe('some-other-answer')
     })
 
+    it('should redirect to exit page even if redirect uri is set', async () => {
```
joseluisgraa, line 161: If we moved the logic to the next page helper function we could test the redirect logic (the result) in there instead of here maybe?

src/server/common/model/section/section-model/section-model-updated.test.js
```diff
@@ -67,10 +86,12 @@ describe('SectionModel.validate', () => {
...
 
   // Reason: We have not finalised how exit pages will behave
-  it.skip('should return ... invalid ? ... if the section hits an exit condition before its complete', () => {
+  it('should return invalid if the section hits an exit condition before its complete', () => {
     const origin = Origin.fromState(exitState)
+    const { isValid, firstInvalidPage } = origin.validate()
 
-    expect(origin.validate().isValid).toBe(true)
+    expect(isValid).toBe(false)
+    expect(firstInvalidPage).toBeInstanceOf(OnOffFarmPage)
```
joseluisgraa, line 94: I guess we need to try and find a way to test the section model without references to specific pages like "OnOffFarmPage" at some point

src/server/origin/summary/controller.js
```diff
@@ -0,0 +1,51 @@
+import { Origin } from '../../common/model/section/origin.js'
+const indexView = 'origin/summary/index.njk'
+export const pageTitle =
+  'Check your answers before you continue your application'
+export const heading = pageTitle
+
+/**
+ * @satisfies {Partial<ServerRoute>}
+ */
+export const originSummaryGetController = {
```
joseluisgraa, line 10: I think this file is not needed as we are now using SummaryPageController

src/server/common/controller/question-page-controller.js
hughfdjackson: But that page doesn't (yet) use the concept of a Page

src/server/common/model/section/section-model/section-model-updated.test.js
hughfdjackson: Yes, I think so.  We don't _yet_ have a great answer to this, but I'm hoping to wait until we've got a bit more experience with this current implementation.

src/server/origin/summary/controller.js
hughfdjackson: removed - good catch :)

joseluisgraa (approved):

PR #74, opened 2024-12-16
DSFAAP-511: Destination type page

joseluisgraa:


src/server/common/model/answer/destination-type.js
```diff
@@ -0,0 +1,74 @@
...
+    } else if (destinationType === 'dedicated-sale') {
+      return 'Dedicated sale for TB (orange market)'
+    } else if (destinationType === 'afu') {
+      return 'Approved finishing unit (AFU)'
+    } else if (destinationType === 'other') {
+      return 'Another destination'
+    } else {
+      return ''
+    }
+  }
```
hughfdjackson, line 52: No pressure to do this (or even do it now), but an object lookup can be nice here: 

```javascript
const destinationTypeMapping = { 
  'slaughter': 'Slaughter',
  'dedicated-sale': 'Dedicated sale for TB (orange market)',
   // etc
}

return destinationTypeMapping[destinationType] ?? ''
```

src/server/destination/exit-page/index.js
```diff
@@ -0,0 +1,6 @@
+import { Page } from '../../common/model/page/page-model.js'
+
+export class DestinationExitPage extends Page {
+  urlPath = '/destination/can-not-use-service'
```
hughfdjackson, line 4: Is this right?  I was expecting 'cannot-user-service' (based on spelling)

src/server/task-list/controller.js
```diff
@@ -32,9 +34,11 @@ export const taskListGetController = {
 
     const destinationGdsTask = buildGdsTaskItem({
       title: 'Movement destination',
-      initialLink: '#',
-      summaryLink: '#',
-      isValid: destination.validate().isValid,
+      initialLink:
+        destinationValidity.firstInvalidPage?.urlPath ??
+        destination.firstPage.urlPath,
```
hughfdjackson, line 39: Hrm - I can see this replicates the pattern we've gone with above, but I *believe* this isn't the pattern we went for.  It was either the firstPage, or it was the summary page.

hughfdjackson (changes requested):
The change requested is on the URL.  

I don't think the task logic is correct, but I don't expect that to be fixed in this PR.

src/server/destination/exit-page/index.js
joseluisgraa: That is the URL in the prototype

src/server/common/model/answer/destination-type.js
joseluisgraa: Nice one I'll update it

src/server/destination/exit-page/index.js
hughfdjackson: okay, let's go with that for now.  It looks like we'll be doing a url sweep later.

hughfdjackson (approved):

user-journey-tests/specs/destination/destinationSelection.spec.js
```diff
@@ -0,0 +1,45 @@
...
+      destinationSelectionPage.pagePath,
+      destinationSelectionPage.pageTitle
+    )
+  })
+
+  it('Should verify that the destination page errors when no option is selected', async () => {
+    await destinationSelectionPage.destinationSelectionErrorTest()
+  })
+
+  it('Should select on the farm radio and continue', async () => {
```
joseluisgraa, line 22: on the farm?

PR #76, opened 2024-12-16
check answers update

eoin-corr-git:


user-journey-tests/helpers/testHelpers/checkAnswers.js
```diff
@@ -87,3 +89,25 @@ export const validateAndAdjustEmail = async (
...
 }
+
+export const validateOnFarmErrorHandling = async (
+  changeElement,
+  checkAnswersHeading,
+  final = true
+) => {
+  await selectElement(changeElement)
+  await toFromFarmPage.selectOnFarmAndContinue()
+  await exitPage.verifyPageHeadingAndTitle(exitPage.pageHeading)
```
DrogoNevets, line 100: does this function not know its own title?

user-journey-tests/specs/finalAnswers.spec.js
```diff
@@ -155,4 +156,16 @@ describe('Check your final answers test', () => {
     )
     await finalAnswersPage.submissionErrorTest()
   })
+
+  it.only('Should verify changing the value to on the farm and navigating back', async () => {
```
DrogoNevets, line 160: only this test?

DrogoNevets (changes requested):

user-journey-tests/helpers/testHelpers/checkAnswers.js
eoin-corr-git: Nah you have to sub it in, that's just calling the method from page.js

hughfdjackson: @eoin-corr-git  - we could do a bit of a refactor here. 

It used to be that each page had a different field for the heading & title, but now we've normalised it, you can call `this.pageTitle` or `this.pageHeading` in the page base class.

hughfdjackson: But I think that's independent of this PR, because that code is everywhere ;)

user-journey-tests/specs/finalAnswers.spec.js
hughfdjackson: :D good catch from Steve there

hughfdjackson (changes requested):
Change the 'test.only'

hughfdjackson (approved):

PR #78, opened 2024-12-17
DSFAAP-607_start-page-updates

joseluisgraa:


src/server/home/index.njk
```diff
@@ -1,42 +1,123 @@
-{% extends 'layouts/page.njk' %}
+{% extends 'layouts/page.njk' %} {% block content %}
```
hughfdjackson, line 1: presumably we should put the content block on a new line? 

Did prettifier recommend this?  :thinking:

src/server/home/index.njk
```diff
@@ -1,42 +1,123 @@
...
+          xmlns="http://www.w3.org/2000/svg"
+          width="17.5"
+          height="19"
+          viewBox="0 0 33 40"
+          focusable="false"
+          aria-hidden="true"
+        >
+          <path fill="currentColor" d="M0 0h13l20 20-20 20H0l20-20z"></path>
+        </svg>
+      </a>
```
hughfdjackson, line 118: @joseluisgraa  - worth using the standard gov uk design system templates with this: 

```jinja2
{% from "govuk/components/button/macro.njk" import govukButton %}

{{ govukButton({
  text: "Start now",
  href: "#",
  isStartButton: true
}) }}
```

hughfdjackson: https://design-system.service.gov.uk/components/button/ <- from here. 

The reason being the usual - we get all of the accessibility benefits out of the box, and don't have to worry about maintaining it as we upgrade the design system.

hughfdjackson (changes requested):

src/server/home/index.njk
joseluisgraa: yes it has been done automatically by prettier, not sure why, and not sure how to override these decisions prettier makes

hughfdjackson (approved):

PR #82, opened 2024-12-18
BAU: refactor question pages to return full payload needed for check-answers page

hughfdjackson:
I want to take this step for a few reasons

## Simplifying Section definition

Currently, each controller is expected to have a getter for every answer - so that it can be rendered in a check-answers summary. 
What those check-answers pages really need are the pages & answers of the relevant summaries. 

Now they only need to fetch questionPageAnswers for all the data needed to render a summary correctly.

## Opens the door to further simplifications

Each SectionModel currently defines the list of questions that are relevant in 'fromState'.  It also dynamically walks the graph of questions (starting from the firstPage).

If we depend only on questionPageAnswers to render check-summary pages, it makes the step of dynamically generating the pages & answers relevant to section from the first page much simpler.

At that stage, a section definition becomes: 

```javascript
export class OriginSection extends SectionModel {
  firstPage = onOffFarmPage
}
```

with no need to update the section model every time we add a new question to the section.  If it's available via nextPage iteration from the section's firstPage, it's in the section automatically - in both the validity checks & in the check-answers summaries.

### DIFF
--- a/src/server/common/model/section/section-model/section-model.js
+++ b/src/server/common/model/section/section-model/section-model.js
@@ -62,7 +62,7 @@ export class SectionModel {
    * @returns {PageAnswer[]}
    */
   get questionPageAnswers() {
-    return this.questionPages.map((page) => ({
+    return this._questionPages.map((page) => ({
       page,
       answer: this._data[page.questionKey].answer
     }))
@@ -71,7 +71,7 @@ export class SectionModel {
   /**
    * returns {QuestionPage[]}
    */
-  get questionPages() {
+  get _questionPages() {
     return this._pages.filter((p) => p instanceof QuestionPage)
   }
 
@@ -84,7 +84,7 @@ export class SectionModel {
     }
 
     if (page instanceof ExitPage) {
-      const questionPages = this.questionPages
+      const questionPages = this._questionPages
       return {
         isValid: false,
         firstInvalidPage: questionPages[questionPages.length - 1]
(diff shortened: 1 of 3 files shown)


src/server/common/model/section/section-model/section-model.js
```diff
@@ -74,10 +79,11 @@ export class SectionModel {
     }
 
     if (page instanceof ExitPage) {
-      const questionPages = this.questionPages
+      const questionPageAnswers = this.questionPageAnswers
       return {
         isValid: false,
-        firstInvalidPage: questionPages[questionPages.length - 1]
+        firstInvalidPage:
+          questionPageAnswers[questionPageAnswers.length - 1].page
```
DrogoNevets, line 86: whilst journey doesnt support this, based on this code, it is possible `this.questionPageAnswers.length` will be zero - so doing `.length -1` without any checking could be problematic

DrogoNevets (changes requested):

src/server/common/model/section/section-model/section-model.js
hughfdjackson: The only way I can imagine to get this error is to do an explicit cast on a page to a QuestionPage:

```javascript
    class InvalidTestSection extends SectionModel {
      firstPage = /** @type QuestionPage */ (new Page()) // not a question page, as required
    }
```

Otherwise, we get an error from typescript.

The reason this guarantees a non-empty array is: 
- _pages pushes the firstPage first
- questionPageAnswers only filters on `instanceof QuestionPage`

So long as firstPage is a QuestionPage, then we won't run into this issue.

PR #84, opened 2024-12-18
[599] Baic page model updates and general licence basic page

DrogoNevets:


src/server/common/controller/page-controller/page-controller.js
```diff
@@ -0,0 +1,61 @@
...
+
+  postHandler(req, h) {
+    const payload = /** @type {NextPage} */ (req.payload)
+    const nextPage = this.page.nextPage()
+
+    if (nextPage instanceof ExitPage) {
+      return h.redirect(nextPage.urlPath)
+    } else {
+      return h.redirect(calculateNextPage(payload.nextPage, nextPage.urlPath))
+    }
```
hughfdjackson, line 59: Hrm - I'm guessing for this ticket, it's not going to be a POST - but a straightforward GET?

I'm not 100% sure which page type we'd use this in.

src/server/common/controller/page-controller/page-controller.js
```diff
@@ -0,0 +1,61 @@
...
+      }
+    }
+  }
+
+  getHandler(req, h) {
+    return h.view(this.page.view, {
+      nextPage: req.query.redirect_uri,
+      pageTitle: this.page.title,
+      heading: this.page.heading,
+      hideQuestion: true,
```
hughfdjackson, line 46: I don't mind merging like this, but it makes more sense (to me) to have a separate non-question content layout for generic pages.

src/server/common/controller/page-controller/page-controller.test.js
```diff
@@ -0,0 +1,101 @@
...
+          },
+          {
+            Cookie: session.sessionID
+          }
+        )
+      )
+
+      const document = parseDocument(payload)
+      expect(statusCode).toBe(statusCodes.ok)
+      expect(document.title).toBe(title)
```
hughfdjackson, line 81: document heading too?

src/server/common/controller/page-controller/page-controller.test.js
```diff
@@ -0,0 +1,101 @@
...
+    })
+
+    it('should redirect to next page, storing question state & preserving the rest of the section state', async () => {
```
hughfdjackson, line 84: doesn't store question state, so probably just 'redirect user to next page'.  Although - as above - I expect we won't use this handler, so I would be in favour of removing it.

src/server/common/model/page/page-model.js
```diff
@@ -7,4 +10,40 @@ export class Page {
...
+
+  /** @returns {string} */
+  get heading() {
+    return this.pageHeading
+  }
+
+  /** @returns {string} */
+  get title() {
+    return this.pageTitle
+  }
```
hughfdjackson, line 34: We could remove these & set the properties to the names we want to access them under?

src/server/common/model/page/page-model.js
```diff
@@ -7,4 +10,40 @@ export class Page {
...
+
+  /** @returns {Record<string, unknown>} */
+  get viewProps() {
+    return {}
+  }
```
hughfdjackson, line 39: I think I like this!

src/server/common/controller/page-controller/page-controller.js
```diff
@@ -0,0 +1,61 @@
...
+            }
+          ])
+        }
+      }
+    }
+  }
+
+  getHandler(req, h) {
+    return h.view(this.page.view, {
+      nextPage: req.query.redirect_uri,
```
hughfdjackson, line 43: Since we're rendering this as a url, we need to call nextPage here, rather than in the POST handler, for content pages.

hughfdjackson: I noticed we've supplemented this with continueURL - I guess removing nextPage is the move in this case?

src/server/common/controller/page-controller/page-controller.js
hughfdjackson: As discussed on slack, we can add a hook in here for an optional POST controller.  I don't think that we should add in a POST controller if it's not needed - since:

- it's a bit misleading to the viewer of the code - e.g. they'll think it's being used
- it's adding something to the customer-facing journey (e.g. a member of the public can post against the server), even though it doesn't make much sense for this use case.

DrogoNevets: as mentioned - this is for the interstitial page found [here](https://applicationsandpermissions-dev-ffeda85e15b1.herokuapp.com/v1_1_0/create-application/contact-and-updates/licence-select-post-can-not-use-this-service) which will need a post handler - conditional could work though.....

src/server/common/controller/page-controller/page-controller.js
DrogoNevets: I think if it gets any more complicated i would agree, but this is purely to hide the default "continue button" that is always present really

the other way was to override the block with an empty slot

src/server/common/model/page/page-model.js
DrogoNevets: to typescript complains as it is a getter somewhere (didnt want to dig around and risk breaking the wider app) and changing it to a property was what it didnt like, hence the slight doubling up

src/server/common/model/page/page-model.js
DrogoNevets: so do i which is why i left it - it was attempt #1 of getting the continue button to look like a link, but there is no GDS style i could find for it, but liked it so left it

src/server/common/controller/page-controller/page-controller.js
DrogoNevets: more than happy to remove - but thought it might be used at some point so better to leave in over remove and put back in again - your call

src/server/common/controller/page-controller/page-controller.test.js
DrogoNevets: done in next push

joseluisgraa (approved):

PR #104, opened 2025-01-03
[DSFAAP-629] add destination to summary page

DrogoNevets:
- **startings**
- **now works**
- **added destination to page**
- **redirect working with constructed url**


src/server/common/model/answer/confirmation/confirmation.js
```diff
@@ -22,10 +22,29 @@ const ensureArray = (value) => (Array.isArray(value) ? value : [value])
...
+      (Array.isArray(this._data?.confirmation) &&
+        this._data?.confirmation?.includes('other')) ||
+      this._data?.confirmation === 'other'
+    ) {
+      value.other = true
+    } else {
+      value.other = false
+    }
+
+    return value
```
hughfdjackson, line 47: I think it'd be simpler to coerce to an array & then treat it consistently, if I'm honest.  This is a little harder to follow for me, and more places where bugs could be introduced.

hughfdjackson (approved):
Minor comment about handling the array in confirmation page, but otherwise approved.

PR #113, opened 2025-01-08
Dsfaap 515 confirmation email

joseluisgraa:


src/server/common/helpers/token/token-utils.js
```diff
@@ -0,0 +1,22 @@
...
+})
+
+export function createToken(apiKey) {
+  const iat = Math.round(Date.now() / 1000)
+  const { iss, secret } = secrets(apiKey)
+
+  return token.generate({ iss, iat }, secret, {
+    header: { typ: 'JWT', alg: 'HS256' }
+  })
+}
```
hughfdjackson, line 22: This token util is (largely) specific to the way that notify handles its tokens, so should be in the same folder with the connector IMO.  Namespace it with notify.

src/server/common/helpers/token/token-utils.test.js
```diff
@@ -0,0 +1,42 @@
...
+    'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789'
+  const secret = apiKey.substring(apiKey.length - 36, apiKey.length)
+  const iss = apiKey.substring(apiKey.length - 73, apiKey.length - 37)
+
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('should generate a token with correct payload and options', () => {
+    const iat = Math.round(Date.now() / 1000)
```
hughfdjackson, line 21: This'll leave our tests with a chance of intermittent failure, since the next time Date.now() is called (inside the function under test), the date may have ticked over. 

Mock?

src/server/common/helpers/token/token-utils.test.js
```diff
@@ -0,0 +1,42 @@
...
+  token: {
+    generate: jest.fn()
+  }
+}))
+
+describe('token-utils', () => {
+  const apiKey =
+    'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789'
+  const secret = apiKey.substring(apiKey.length - 36, apiKey.length)
+  const iss = apiKey.substring(apiKey.length - 73, apiKey.length - 37)
```
hughfdjackson, line 14: For readability, I think we could steal from the gov notify docs here: 

image

So:

```javascript
  const apiKey =
    'my_test_key-26785a09-ab16-4eb0-8407-a37497a57506-3d844edf-8d35-48ac-975b-e847b4f122b0'
  const secret = '3d844edf-8d35-48ac-975b-e847b4f122b0'
  const iss = '26785a09-ab16-4eb0-8407-a37497a57506'
  ```
  
  This also stops us from using the core logic of the function under test within the test - e.g. if we got a bug with where we set our boundaries & we repeat it in the test & implementation, then the test-as-is won't catch this. 
  
 If we're being specific about what we expect to extract, then the test will.

File: src/server/submit/confirmation/index.js [hunk 4]
```diff
@@ -0,0 +1,28 @@
...
+      server.route([
+        {
+          method: 'GET',
+          path: '/submit/confirmation',
+          ...submitConfirmationController
+        }
+      ])
+    }
+  }
+}
```
hughfdjackson, line 24: We've got a generic page controller for this now - that'll generate the plugin for us. See `src/server/destination/general-licence/index.js` for an example

hughfdjackson (changes requested):

hughfdjackson:
I also believe we're missing an update to the readme for the new environment configuration

src/server/check-answers/index.test.js
```diff
@@ -15,69 +68,198 @@ describe('#checkAnswers', () => {
...
+        },
+        {
+          Cookie: session.sessionID
+        }
+      )
+    )
+
+    expect(statusCode).toBe(statusCodes.redirect)
+    expect(headers.location).toBe(taskListIncompleteUri)
+  })
```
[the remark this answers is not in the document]
joseluisgraa: removed, good spot

src/server/check-answers/index.test.js
```diff
@@ -15,69 +68,198 @@ describe('#checkAnswers', () => {
...
+
+  it('Should not send email and display an error', async () => {
```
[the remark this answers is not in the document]
joseluisgraa: updated

src/server/common/connectors/notify/notify-token-utils.test.js
```diff
@@ -17,7 +17,8 @@ describe('token-utils', () => {
   })
 
   it('should generate a token with correct payload and options', () => {
-    const iat = Math.round(Date.now() / 1000)
+    const iat = 1
+    jest.spyOn(Math, 'round').mockReturnValue(iat)
```
hughfdjackson, line 21: Sorry, I was really imprecise with my original comment.  It'd be better to mock Date.now(), since that's what we want to prove the function does - take the current time, device by 1000, round down and use as appropriate in the JWT token.

src/server/submit/confirmation/index.js
```diff
@@ -1,28 +1,23 @@
...
-        {
-          method: 'GET',
-          path: '/submit/confirmation',
-          ...submitConfirmationController
-        }
-      ])
-    }
-  }
+  pageTitle = 'Your animal disease movement licence application'
+  pageHeading = 'Your animal disease movement licence application'
```
hughfdjackson, line 15: Amazed sonarcloud didn't give us a little stick for this - we've been extracting these into a const just to be sure they're identical, where they ought to be.

hughfdjackson (changes requested):

src/server/submit/confirmation/index.js
joseluisgraa: there are loads of instances where we have the same two values defined like this but I'll change this one at least

src/server/common/connectors/notify/notify-token-utils.test.js
joseluisgraa: done

hughfdjackson (approved):

hughfdjackson (approved):

PR #119, opened 2025-01-14
[706] start page title

DrogoNevets:


src/server/home/index.njk
```diff
@@ -103,11 +103,31 @@
           choose the option for TB licencing. Lines are open from Monday to
           Friday between 8.30am and 5pm.
         </p>
-
+        <<<<<<< Updated upstream
```
joseluisgraa, line 106: Something is not right here

src/server/home/controller.js
```diff
@@ -6,7 +8,7 @@
 export const homeController = {
   handler(_request, h) {
     return h.view('home/index', {
-      pageTitle: 'Apply for a Bovine Tuberculosis (TB) movement licence',
+      pageTitle: config.get('serviceName'),
       heading: 'Apply for a Bovine Tuberculosis (TB) movement licence'
```
joseluisgraa, line 12: Do we still need this property then?

src/server/home/controller.js
```diff
@@ -6,7 +8,7 @@
 export const homeController = {
   handler(_request, h) {
     return h.view('home/index', {
-      pageTitle: 'Apply for a Bovine Tuberculosis (TB) movement licence',
+      pageTitle: config.get('serviceName'),
```
joseluisgraa, line 11: Is it going to change between envs? If not we could just set the actual value of the title here.

src/server/home/controller.js
joseluisgraa: Does the title tag not need to be changed too? @hughfdjackson ?

src/server/home/controller.js
hughfdjackson: This is one of those cases where we want something to be invarient - no matter the environment. 

If we have it in config, it weakens our tests (e.g. sure it's checked the service in the publish checker - but what test says that's the service name in production?).

IMO, this is definitely best to left hardcoded - either directly, or in a `const` somewhere in the code.

src/server/home/controller.js
hughfdjackson: It does - that's the second AC in the ticket related to this PR

src/server/home/controller.js
DrogoNevets: to be clear you are asking be to go through everywhere the title config prop is used and hard code it? totally fine, just making sure i understand the ask

joseluisgraa (approved):

PR #120, opened 2025-01-14
[707] feedback link added to confirmation page

DrogoNevets:
- **use service name on start page**
- **add feedback to confirmation page**

### DIFF
--- a/src/server/home/controller.test.js
+++ b/src/server/home/controller.test.js
@@ -1,6 +1,7 @@
 import { createServer } from '~/src/server/index.js'
 import { statusCodes } from '~/src/server/common/constants/status-codes.js'
 import { parseDocument } from '~/src/server/common/test-helpers/dom.js'
+import { config } from '~/src/config/config.js'
 
 describe('#homeController', () => {
   /** @type {Server} */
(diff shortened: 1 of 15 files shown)


src/server/home/index.njk
```diff
@@ -103,11 +103,31 @@
           choose the option for TB licencing. Lines are open from Monday to
           Friday between 8.30am and 5pm.
         </p>
-
+        <<<<<<< Updated upstream
```
joseluisgraa, line 106: Need to review this

src/server/home/controller.test.js
```diff
@@ -21,11 +21,8 @@ describe('#homeController', () => {
       url: '/'
     })
 
-    expect(parseDocument(payload).title).toBe(
-      'Apply for a Bovine Tuberculosis (TB) movement licence'
-    )
-
     expect(statusCode).toBe(statusCodes.ok)
+    expect(parseDocument(payload).title).toBe('Applications and permissions')
```
joseluisgraa, line 25: I don't think this is the correct title we need to set

joseluisgraa (approved):

PR #126, opened 2025-01-16
DSFAAP-734: remove continue button

hughfdjackson:
We're using the question layout for our content page - which is probably the wrong move.

Also adds content coverage for this page via snapshot tests.  It's the absence of this coverage that caused this regression to be possible without us noticing.


DrogoNevets (approved):

hughfdjackson:
Just for historical (I probably ought to have put this in the comment for the last commit): 

I tried quite a lot of options to get the pipeline tests to pass - the crux is that the npm run server:build uses babel over to copy over files to the .server but*excludes* test files.   However, it includes the snapshots in the.server copy.

I tried a lot of things to fix this:

- add an ignore to babel command in npm run build:server: doesn't work
  because Babel doesn't ignore files it doesn't consider javascript (and
  test.js.snap is not js according to Babel)
- change the test.js.snap suffix to test.snap.js in jest: no option I
  can find
- inline the tests: can't use inline snapshot tests with prettifier 3 &
  jest
- get Babel to consider test.js.snap as javascript: no option I can find

So the compromise is that we've got snapshot files in our server build - which isn't the cleanest, but doesn't impact on the functionality in any visible way.

PR #128, opened 2025-01-16
DSFAAP-731: Added help text to page template

joseluisgraa:


hughfdjackson (approved):

user-journey-tests/page-objects/page.js
```diff
@@ -72,9 +88,29 @@ class Page {
...
+    )
+    await page.validateElementVisibleAndText(
+      this.getCookiesFooterLink(),
+      'Cookies'
+    )
+    await page.validateElementVisibleAndText(
+      this.getAccessibilityFooterLink(),
+      'Accessibility statement'
+    )
+  }
```
hughfdjackson, line 108: Can we also verify the hrefs point to the site as expected?  

I know we've tested that we can actually physically navigate to the destination page on the landing page.  That's definitely a good test to do once, but too expensive to do on every page. 

But validating the href would give us a lot of confidence without a lot of extra execution time & no change of page.

user-journey-tests/helpers/page.js
```diff
@@ -84,6 +84,11 @@ export const validateElementVisibleAndText = async (element, text) => {
   }
 }
 
+export const validateHrefOfElement = async (element, href) => {
+  const hrefValue = await element.getAttribute('href')
+  expect(hrefValue).toContain(href)
```
hughfdjackson, line 89: This one we want to be 'toBe'

hughfdjackson (changes requested):

PR #150, opened 2025-01-23
[744] record errors

DrogoNevets:


src/server/common/controller/generic-page-controller/index.js
```diff
@@ -27,7 +23,7 @@ import { NotImplementedError } from '../../helpers/not-implemented-error.js'
  */
 
 export default class GenericPageController {
-  metrics = createMetricsLogger()
+  logger = createLogger()
```
hughfdjackson, line 26: This is the line that caused the issues last time. 😰 

I believe the issue is that every instance of this will create its own logger (and in the unit testing context, that happens a lot).

Do we have another approach we could use (e.g. [instantiate the logger outside of the class, and spawn children](https://github.com/pinojs/pino/blob/main/docs/child-loggers.md)?) that would avoid creating multiple loggers?

Also, do you see the memory leak warnings when you run your tests with this change in?

src/server/common/controller/generic-page-controller/index.js
```diff
@@ -50,6 +46,16 @@ export default class GenericPageController {
     return result
   }
 
+  /**
+   *
+   * @param {import('../../model/answer/validation.js').AnswerErrors} errors
+   */
+  recordErrors(errors) {
+    Object.entries(errors).forEach(([key, value]) => {
+      this.sendErrorMetric(key, value.text)
```
hughfdjackson, line 55: Not a metric anymore - `logError`?

hughfdjackson (changes requested):

src/server/common/controller/generic-page-controller/index.js
```diff
@@ -63,13 +69,21 @@ export default class GenericPageController {
...
+
+  /**
+   *
+   * @param {string} field
+   * @param {string} error
+   */
+  sendErrorMetric(field, error) {
+    if (!config.get('isProduction')) {
+      return
     }
```
hughfdjackson, line 84: We can output these all the times - they're just logs - and it'd be useful to!

src/server/common/controller/generic-page-controller/index.js
```diff
@@ -63,13 +69,21 @@ export default class GenericPageController {
...
+
+    this.logger.info(`${field} errored:${error}-${this.page.urlPath}`)
```
hughfdjackson, line 86: Could we make this more human readable? 

e.g. `User encountered a validation error on ${this.page.urlPath}, on the ${field} field : ${error}`?

src/server/common/controller/page-controller/page-controller.js
```diff
@@ -49,6 +49,7 @@ export class PageController extends GenericPageController {
 
         /** @param {Server} server */
         register: (server) => {
+          this.server = server
```
hughfdjackson, line 52: Is this used anywhere?

src/server/common/controller/generic-page-controller/index.js
DrogoNevets: no memory leak warnings when i run tests no

src/server/common/controller/question-page-controller/question-page-controller.test.js
```diff
@@ -296,6 +297,9 @@ describe('QuestionPageController', () => {
       })
 
       it('Should display an error and set next page appropriately', async () => {
+        config.set('isProduction', true)
```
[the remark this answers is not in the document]
DrogoNevets: not needed now anyways

hughfdjackson (approved):

PR #156, opened 2025-01-24
[743] add logging to start journey

DrogoNevets:


src/server/common/test-helpers/csrf.js
```diff
@@ -79,7 +79,8 @@ export const withCsrfProtection = (injectionOptions, headers = {}) => {
     headers: {
       ...(injectionOptions.headers ?? {}),
       ...headers,
-      Cookie: ['crumb=csrf-value', headers?.Cookie].filter((v) => v).join(';')
+      Cookie: ['crumb=csrf-value', headers?.Cookie].filter((v) => v).join(';'),
+      Referer: headers.Referer ?? ''
```
hughfdjackson, line 83: Is this necessary - since if we include it in `headers`, it'll be available here? 

If this is breaking one of our tests (e.g. for taskListGetController), then I think that implies that the controller is not robust enough for a non-essential header being missing rather than we should make sure it's always there

src/server/task-list/controller.test.js
```diff
@@ -113,6 +123,77 @@ describe('#taskListController', () => {
...
+          Referer: `https://${server.info.host}/` // simulate the index page of a application
+        }
+      )
+    )
+
+    expect(statusCode).toBe(statusCodes.ok)
+    expect(mockHapiLoggerInfo).toHaveBeenCalledTimes(1)
+  })
+
+  it('should log that the user came from any other page', async () => {
```
hughfdjackson, line 144: `should log` -> `should not log`?

src/server/task-list/controller.test.js
```diff
@@ -113,6 +123,77 @@ describe('#taskListController', () => {
...
+          Referer: `https://star-trek.com/` // simulate the index page of a application
+        }
+      )
+    )
+
+    expect(statusCode).toBe(statusCodes.ok)
+    expect(mockHapiLoggerInfo).toHaveBeenCalledTimes(0)
+  })
+
+  it('should not log that the user came from start page', async () => {
```
hughfdjackson, line 180: `should not log that the user came directly to the task list page`?

hughfdjackson (changes requested):

PR #162, opened 2025-01-27
[WDIO] dynamic thread count

DrogoNevets:


user-journey-tests/config/browserstackMetrics.js
```diff
@@ -0,0 +1,41 @@
...
+
+    const data = await response.json()
+    const {
+      parallel_sessions_max_allowed: threadsAllowed,
+      parallel_sessions_running: threadsRunning
+    } = data
+    const availableThreads = threadsAllowed - threadsRunning
+    const threadsToUse = Math.floor(availableThreads / 2)
+
+    if (threadsToUse < defaultMinimum && threadsToUse < availableThreads) {
```
hughfdjackson, line 29: I'm not sure if it's just my eyesight, but this doesn't read quite right to me. 

Threads to use will always be < available threads, because it's calculated as `const threadsToUse = Math.floor(availableThreads / 2)`

It looks to me that we want: 

```
if (defaultMinimum > availableThreads) {
  return availableThreads
} else {
  return defaultMinimum
}
```

Also, isn't it more of a defaultMaximum?

DrogoNevets: can you tell im working on 3hrs sleep?

hughfdjackson (approved):
As per slack, I've run out of brain-steam.  I don't want to stand in the way of trying something like this, so I think it's worth merging & refining later if needs be.

PR #169, opened 2025-01-29
DSFAAP-799: port origin address to address template

hughfdjackson:


src/server/common/model/answer/address/address.js
```diff
@@ -127,4 +127,14 @@ export class AddressAnswer extends AnswerModel {
   static fromState(state) {
     return new AddressAnswer(state)
   }
+
+  viewModel({ validate }) {
```
DrogoNevets, line 131: missing typedefs
suggest we group class properties some how, eg all static functions together, getters/setters together, etc

DrogoNevets (changes requested):

PR #173, opened 2025-01-29
Dsfaap 785 v2 separate incoming cattle page

joseluisgraa:


DrogoNevets (approved):

src/server/biosecurity/kept-separately/index.js
```diff
@@ -0,0 +1,38 @@
...
+
+export class KeptSeparatelyPage extends QuestionPage {
+  urlPath = '/biosecurity/kept-separately'
+  sectionKey = 'biosecurity'
+  question = 'Will you separate the incoming cattle from the resident herd?'
+  questionKey = 'keptSeparately'
+  view = 'biosecurity/kept-separately/index'
+  Answer = KeptSeparatelyAnswer
+
+  /** @param {AnswerModel} answer */
```
hughfdjackson, line 17: We can be more specific to {KeptSeparatelyAnswer} here, I believe.

src/server/common/model/answer/radio-button/radio-button.test.js
```diff
@@ -168,14 +169,39 @@ describe('RadioButton', () => {
...
         errorMessage: { text: 'Select an option' }
       })
     })
+
+    describe('radio button layout', () => {
+      const originalLayout = testRadioConfig.layout
+
+      afterEach(() => {
+        testRadioConfig.layout = originalLayout
+      })
```
hughfdjackson, line 189: I think it's much clearer here if we declare a new class - rather than the user having to keep track of how the RadioButtonTest config has been mutated.

With the double-configs, I can see why this pattern is tempting, but I'd rather:
-  port us over to [this pattern](https://github.com/DEFRA/apha-apps-perms-move-animal-ui/pull/171/files#diff-ee1323f5f98f8878ed19567c758f80fbed2dcd4504e6cd2214f6cfb4e45a1c0dR58-R69) for RadioButton 
- set up testing variations, e.g. [like this](https://github.com/DEFRA/apha-apps-perms-move-animal-ui/pull/171/files#diff-14aa05114e7d09fc8ba7bd396d47afa3f1400eba67640cf3fbbdfcaaa8f9e3bcR17-R22)

Would you mind to do this as part of this ticket?  At the end of DSFAAP-799, I'll make sure all of the models are using this simpler config / static config set up.

src/server/common/model/answer/kept-separately/kept-separately.js
```diff
@@ -0,0 +1,31 @@
...
+}
+
+/**
+ * export @typedef {'yes'|'no'} KeptSeparatelyData
+ * @typedef {{ keptSeparately: KeptSeparatelyData }} KeptSeparatelyPayload
+ */
+
+/** @augments {RadioButtonAnswer<KeptSeparatelyPayload>} */
+export class KeptSeparatelyAnswer extends RadioButtonAnswer {
+  get config() {
```
hughfdjackson, line 24: Once we've updated the config strategy, we can replace this with:

```
export class KeptSeparatelyAnswer extends RadioButtonAnswer {
  static config = keptSeparatelyConfig
}
```

(or just inline the config, as we have done elsewhere)

user-journey-tests/specs/biosecurity/keptSeparatelySelection.spec.js
```diff
@@ -0,0 +1,37 @@
...
+  it('Should select Yes, continue and check its maintained', async () => {
+    await keptSeparatelyPage.selectYesAndContinue()
+
+    await expect(keptSeparatelyPage.pageError).not.toBeDisplayed()
+    await expect(keptSeparatelyPage.errorSummary).not.toBeDisplayed()
+    await waitForPagePath(grazingPage.pagePath)
+
+    await grazingPage.selectBackLink()
+    await waitForPagePath(keptSeparatelyPage.pagePath)
+    await expect(keptSeparatelyPage.yesRadio).toBeSelected()
```
hughfdjackson, line 27: I think that the browsers typically play this out of cache, so I don't think this test will tell us if the server is maintaining state. 

At this point in time, we might have to put an 'await browser.refresh()` in there?

hughfdjackson (dismissed):

src/server/biosecurity/kept-separately/index.test.js
```diff
@@ -0,0 +1,83 @@
...
+    expect(keptSeparately).toHaveProperty('plugin')
+    const plugin = /** @type {PluginBase<void> & PluginNameVersion} */ (
+      keptSeparately.plugin
+    )
+    expect(plugin).toHaveProperty('name')
+    expect(plugin.name).toBe(`${sectionKey}-${questionKey}`)
+    expect(plugin).toHaveProperty('register')
+  })
+
+  describe('#keptSeparatelyPage.content', () => {
```
hughfdjackson, line 74: `describePageSnapshot` already has a describe block embedded in it, so no need for this extra one

src/server/common/model/answer/kept-separately/kept-separately.js
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * export @typedef {'yes'|'no'} KeptSeparatelyData
+ * @typedef {{ keptSeparately: KeptSeparatelyData }} KeptSeparatelyPayload
+ */
+
+/**
+ * @augments {RadioButtonAnswer<KeptSeparatelyPayload>}
+ */
+export class KeptSeparatelyAnswer extends RadioButtonAnswer {
```
hughfdjackson, line 13: We have a 'Yes No' model for this.  If you feel that's overkill, then we can remove the model and go ahead with this version.

PR #180, opened 2025-01-31
Accessibility statement page

joseluisgraa:


src/server/accessibility/__snapshots__/index.test.js.snap
```diff
@@ -0,0 +1,53 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`#accessibilityStatementPage should render expected response and content 1`] = `
+"
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <div class="govuk-body" data-testid="app-page-body">
```
hughfdjackson, line 7: I can't see this extra wrapper anywhere in https://design-system.service.gov.uk/styles/layout/ - should we remove?  and move the `data-testid` up to the element above?

src/server/accessibility/index.js
```diff
@@ -0,0 +1,24 @@
+import { Page } from '~/src/server/common/model/page/page-model.js'
+import { PageController } from '~/src/server/common/controller/page-controller/page-controller.js'
+
+export class AccessibilityStatementPage extends Page {
+  sectionKey = 'policies'
+  key = 'accessibility-statement'
+  pageTitle = 'Accessibility statement'
```
hughfdjackson, line 7: :D one to extract into a 'const' above - otherwise sonarcloud will let us know all about it.

hughfdjackson (changes requested):

src/server/accessibility/__snapshots__/index.test.js.snap
joseluisgraa: everywhere? or just this file?

hughfdjackson: ahh - we've done this everywhere?  let's not worry about that in this ticket, then :)

src/server/accessibility/index.js
joseluisgraa: done

DrogoNevets (approved):

PR #181, opened 2025-01-31
[801] bio sec map

DrogoNevets:
- **start of basic upload**
- **start of file upload**


compose.yml
```diff
@@ -8,25 +7,88 @@ services:
     networks:
       - cdp-tenant
 
-  your-frontend:
-    build: ./
+  # your-frontend:
+  #   build: ./
```
hughfdjackson, line 11: @DrogoNevets  - We're going to need this block for our pr checker to work.  Just flagging it here as a reminder, I know this is WIP.

DrogoNevets:
Next PR will
move the upload stuff into aa proper answer model

DrogoNevets:
### Biosecurity upload happy path only
This PR adds upload of biosecurity map, happy path only

There are known omissions
* doesn't use answer models
* doesnt error check
* doesnt belong to a section
* redirects to a 404
* doesnt test client side code

These will be fixed in subsequent PRs - the next being an answer model for file upload and file meta data

package.json
```diff
@@ -68,12 +68,14 @@
     "nunjucks": "^3.2.4",
     "pino": "^9.4.0",
     "pino-pretty": "^11.2.2",
-    "undici": "^6.19.8"
+    "undici": "^6.19.8",
+    "util": "0.12.5"
```
hughfdjackson, line 72: I don't see us using this dependency in the code below - presumably this was a dependency for frontend tests that we've since removed?

DrogoNevets: no - this was a audit fix thing

sonar-project.properties
```diff
@@ -13,7 +13,7 @@ sonar.links.scm=https://github.com/DEFRA/apha-apps-perms-move-animal-ui
 sonar.links.issue=https://github.com/DEFRA/apha-apps-perms-move-animal-ui/issues
 
 sonar.sources=src/
-sonar.exclusions=src/**/*.test.js,src/server/common/test-helpers/*
+sonar.exclusions=src/**/*.test.js,src/server/common/test-helpers/*,src/client/javascripts/**/*.js
```
hughfdjackson, line 16: Looks like this excludes all client-side code from sonarcloud analysis (not just coverage analysis), which is further than we'd ideally go.

package.json
DrogoNevets: but that mightve been for journey tests - ill test to make sure

DrogoNevets: oh you mean util - yes thats not needed anymore - the comments made it look like it was undici

src/client/javascripts/file-upload-spinner.js
```diff
@@ -0,0 +1,18 @@
+document.addEventListener('DOMContentLoaded', () => {
+  const form = document.querySelector('form')
+  const uploadSpinner = document.querySelector('form ~ .apha_upload-spinner')
```
hughfdjackson, line 3: Is this a bit fragile?  Would expect we could put an id on the spinner & address it directly.

hughfdjackson (changes requested):
changes requested: 

- remove unnecessary node module (if I'm right about that)
- include sonarcloud analysis for clientside code

hughfdjackson (approved):

PR #185, opened 2025-02-03
User journey test fixes

joseluisgraa:


user-journey-tests/helpers/testHelpers/checkAnswers.js
```diff
@@ -137,17 +137,19 @@ export const validateReceiveMethod = async (changeLink, valueElement) => {
   await expect(receiveMethodPage.emailRadio).toBeSelected()
   await receiveMethodPage.selectEmailAndContinue()
 
-  await validateElementVisibleAndText(valueElement, 'email')
+  await validateElementVisibleAndText(valueElement, 'Email')
```
hughfdjackson, line 140: @joseluisgraa  - huh - shouldn't this mean that publish should have been failing since we changed 'email' to 'Email' (which was quite a while ago)?

This looks like the right fix, but it's a mystery to me why we weren't having only total failures without this fix in.

hughfdjackson: I may be answering my own question - we have, and I've not been paying enough attention: https://github.com/DEFRA/apha-apps-perms-move-animal-ui/actions/workflows/publish.yml

joseluisgraa: Yes it is a mistery, there was another place where we seemed to be using a test-id that I couldn't see in the code...

user-journey-tests/page-objects/finalAnswersPage.js
```diff
@@ -77,7 +77,7 @@ class FinalAnswersPage extends Page {
   }
 
   get receiveMethodChange() {
-    return $('[data-testid="licence-choice-change-link"]')
+    return $('[data-testid="receiveMethod-change-link"]')
```
joseluisgraa, line 80: This one @hughfdjackson

user-journey-tests/helpers/testHelpers/finalAnswers.js
```diff
@@ -0,0 +1,49 @@
...
+import finalAnswersPage from '../../page-objects/finalAnswersPage'
+import landingPage from '../../page-objects/landingPage'
+import checkAnswersPage from '../../page-objects/origin/checkAnswersPage'
+import licenceAnswersPage from '../../page-objects/receiving-the-licence/licenceAnswersPage'
+import taskListPage from '../../page-objects/taskListPage'
+import completeDestinationTask from './destination'
+import { completeOriginTaskAnswersCustom } from './movementLicence'
+import { completeLicenceTaskAnswersCustom } from './receivingLicence'
+
+export const completeApplication = async (originObject, licenceObject) => {
```
joseluisgraa, line 11: We probably should allow all sections to be passed in as arguments with some default values if not provided

PR #187, opened 2025-02-04
Dsfaap 781 biosecurity details section

joseluisgraa:


src/server/biosecurity/disinfection/index.js
```diff
@@ -1,6 +1,8 @@
-import { QuestionPage } from '~/src/server/common/model/page/question-page-model.js'
```
hughfdjackson, line 1: `biosecurity` -> `biosecurity-details`? We have another section called `biosecurity-map` incoming, so we'll end up with a weird ambiguity if we don't call it by its full name

src/server/common/model/section/section-model/section-model.js
```diff
@@ -92,4 +96,26 @@ export class SectionModel {
...
     return new this(pages)
   }
+
+  // eslint-disable-next-line jsdoc/require-returns-check
+  /**
+   * @param {Request} req
+   * @returns {object}
+   */
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  buildGdsTaskDetails(req) {
```
hughfdjackson, line 106: If we move the section template next to the section-model & give it the same name (`section-model.njk`), we could happily call this `viewModel` & it'd follow the pattern of the answer page.

just a thought!

src/server/biosecurity/disinfection/index.js
joseluisgraa: I've followed the same convention as in the URLs in the prototype, biosecurity for this one and biosecurity-map for the other. I can rename the stuff but URLs will still be kept as biosecurity. Would you want the section key under which we store the data in this section to also be called biosecurityDetails?

src/server/common/model/section/section-model/section-model.js
joseluisgraa: Do we have a section template? Maybe you are talking about the task list, but we also have the create summary macro so I'm not sure as both are related to sections.  Shall we leave this one as is and maybe have a think after?

src/server/biosecurity/disinfection/index.js
hughfdjackson: ah - that's fine. :) so long as we have a principled way to separate the sections.

src/server/common/model/section/section-model/section-model.js
hughfdjackson: create summary macro is the equivalent of the session-model view (IMO).  In the same way that the input view is not the only place the answer data is used, but it is the main place. 

But for this PR, leave it as is and we can have a think after.

src/server/common/model/application/application.test.js
```diff
@@ -25,20 +30,33 @@ const destinationDefaultState = {
...
 
-    const application = ApplicationModel.fromState(state)
+    const application = ApplicationModel.fromState(request)
 
     expect(application).toBeInstanceOf(ApplicationModel)
 
     expect(application.tasks.origin).toBeInstanceOf(OriginSection)
     expect(application.tasks.licence).toBeInstanceOf(LicenceSection)
     expect(application.tasks.destination).toBeInstanceOf(DestinationSection)
+    expect(application.tasks.biosecurity).toBeInstanceOf(BiosecuritySection)
```
hughfdjackson, line 60: We don't confirm that the state really makes it through to the various sections here - only that we've instantiated them.

hughfdjackson: pushing  up a suggested test case for this in a min

hughfdjackson (approved):

hughfdjackson (approved):

PR #191, opened 2025-02-05
DSFAAP-808: issue warns log for non-server errors

hughfdjackson:


src/server/common/helpers/errors.test.js
```diff
@@ -107,14 +104,40 @@ describe('#catchAll', () => {
...
   })
+
+  const statuses4xx = [
+    statusCodes.notFound,
+    statusCodes.forbidden,
+    statusCodes.unauthorized,
+    statusCodes.imATeapot
+  ].map((status) => [status])
+
+  test.each(statuses4xx)('should call warn logger on 4xx', (status) => {
```
joseluisgraa, line 122: should we use "it" here? same on line 134

hughfdjackson: agreed - will convert the whole file over.

PR #193, opened 2025-02-06
[801] upload uses answer model

DrogoNevets:


DrogoNevets:
Next step is error messages on upload page

src/server/biosecurity-map/upload-progress/__snapshots__/index.test.js.snap
```diff
@@ -19,3 +19,5 @@ exports[`#UploadPlan #UploadProgressPage should render expected response and con
   </div>
         "
 `;
+
+exports[`#UploadPlan valid upload should render the upload progress page 1`] = `""`;
```
joseluisgraa, line 23: is this correct?

src/server/biosecurity-map/upload-config.js
```diff
@@ -0,0 +1,4 @@
+export const uploadConfig = {
+  sectionKey: 'biosecurity-map',
+  questionKey: 'upload-plan'
```
joseluisgraa, line 3: I can see why you want to have the section key defined somewhere to be reused across all pages in the section but the question key is only used in one page so I don't think we need it here?

joseluisgraa: You seem to be using it in two pages, the same question key, is that needed because of the upload progress page?

src/server/biosecurity-map/upload-plan/index.js
```diff
@@ -39,15 +40,28 @@ export class UploadPlanController extends QuestionPageController {
 
     const data = JSON.parse(response.payload.toString())
 
-    req.yar.set('upload', data)
+    const answer = new this.page.Answer({
```
joseluisgraa, line 43: Should this.page not be uploadPlanPage ?

src/server/biosecurity-map/upload-progress/index.js
```diff
@@ -1,45 +1,63 @@
...
+import { UploadPlanPage } from '../upload-plan/index.js'
 
-/**
- * @import {NextPage} from '../../common/helpers/next-page.js'
- * @import {ConfirmationPayload} from '../../common/model/answer/confirmation/confirmation.js'
- */
-
-class BioSecuritySummary extends QuestionPage {
-  urlPath = '/biosecurity/check-answers'
+class BioSecuritySummary extends Page {
```
joseluisgraa, line 10: could move this bit to the actual file that we will use for the summary page?

src/server/biosecurity-map/upload-config.js
DrogoNevets: for file upload its actually used in 2 pages

src/server/biosecurity-map/upload-progress/__snapshots__/index.test.js.snap
DrogoNevets: this is a snapshot and is 100% generated so i guess so

src/server/biosecurity-map/upload-progress/index.js
DrogoNevets: will do in turn

src/server/biosecurity-map/upload-plan/index.js
DrogoNevets: effectively the same due to line 74
new UploadPlanController(uploadPlanPage)

src/server/common/model/answer/biosecurity-map/biosecurity-map.js
```diff
@@ -0,0 +1,105 @@
...
+ *   metadata: {},
+ *   form: {
+ *     crumb: string,
+ *     file: any
+ *   },
+ *   numberOfRejectedFiles: number
+ * }
+ * } | undefined} BiosecurityMapData
+ *
+ * export @typedef {{ metadata: {
```
joseluisgraa, line 23: I think you can use BiosecurityMapData here

src/server/biosecurity-map/upload-config.js
DrogoNevets: the actual upload, and the progress page

src/server/common/model/answer/biosecurity-map/biosecurity-map.js
DrogoNevets: done

PR #211, opened 2025-02-12
BAU: checkbox models

hughfdjackson:


src/server/common/model/answer/confirmation/confirmation.test.js
```diff
@@ -52,70 +52,61 @@ describe('#ConfirmationModel', () => {
   describe('toState', () => {
     it('should return an object with the correct data (confirm only)', () => {
       const confirmation = new ConfirmationAnswer({ confirmation: ['confirm'] })
-      expect(confirmation.toState()).toEqual({ confirmation: ['confirm'] })
+      expect(confirmation.toState()).toEqual(['confirm'])
```
hughfdjackson, line 55: These tests need cutting way back, since they replicate the tests from the base class they inherit from

src/server/common/model/answer/confirmation/confirmation.js
```diff
@@ -1,68 +1,33 @@
...
-   * @returns {ConfirmationData | undefined}
-   */
-  toState() {
-    return {
-      confirmation: ensureArray(this._data?.confirmation ?? [])
+import { CheckboxAnswer } from '../checkbox/checkbox.js'
+/** @import {CheckboxConfig, CheckboxData} from '../checkbox/checkbox.js' */
+const checkboxEmptyError = 'You need to tick a declaration box'
+
+// value: "confirm",
```
joseluisgraa, line 5: To tidy up

src/server/common/model/answer/checkbox/checkbox.test.js
```diff
@@ -0,0 +1,243 @@
+import { CheckboxAnswer } from './checkbox.js'
+
+/** @import {CheckboxConfig} from './checkbox.js' */
+
+const checkboxEmptyError = 'Select at least one checkbox'
+
+/** @type {CheckboxConfig} */
```
joseluisgraa, line 7: We could do it more generic and not use the badger page values, but I can actually do that once I've worked on that page

src/server/common/model/answer/checkbox/checkbox.js
```diff
@@ -0,0 +1,143 @@
...
+
+  return Joi.object({ [config.payloadKey]: optionsSchema })
+}
+
+/**
+ * @template T
+ * @param {T[] | T} value
+ * @returns T[]
+ */
+const ensureArray = (value) => (Array.isArray(value) ? value : [value])
```
joseluisgraa, line 33: could be moved to a helper file?

joseluisgraa: Not sure about that, but we could even allow undefined input and return empty array so we can simplify line 114?

src/server/common/model/answer/checkbox/checkbox.js
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  /**
+   * @param {AnswerViewModelOptions} options
+   */
+  viewModel({ question, validate }) {
+    const values = this.value
+
+    const viewModel = {
+      name: this.config.payloadKey,
```
joseluisgraa, line 85: Should we add hint?

src/server/check-answers/index.njk
```diff
@@ -21,35 +21,11 @@
...
-        attributes: {
-          'data-testid': 'someone-else-checkbox'
-        }
-      }]
-    })
-  }}
 {% endblock %}
+{%
+  set viewModelOptions = {
+  question: 'Before you submit your application, you must select one of the declaration check boxes.',
```
joseluisgraa, line 28: not sure I understand why this is here

src/server/common/model/answer/checkbox/checkbox.js
hughfdjackson: @joseluisgraa  - I'd happily do so, if we discover another use case for it?

This is the only type of form input I know of that results in a Hapi body that has either 'string | string[]', which is the use-case we're handling here. 

:thinking:

hughfdjackson: > Not sure about that, but we could even allow undefined input and return empty array so we can simplify line 114?

This is a good suggestion - updating.

src/server/common/model/answer/checkbox/checkbox.js
hughfdjackson: I've only included things that we currently can cover with a `snapshot` test - I was going to add in a hint when we got to the badger page. 

Otherwise I don't have a great way to test it - but you're right that we need it

src/server/common/model/answer/confirmation/confirmation.js
hughfdjackson: :+1:

src/server/check-answers/index.njk
hughfdjackson: This was a temporary hack to get the user-journey tests to pass.  I've revisited it now :)

joseluisgraa (approved):

PR #227, opened 2025-02-19
DSFAAP-853: Allow dynamic radio button configuration & implement routing for origin type

hughfdjackson:
I've added the option to pass a config factory.  This allows us to:

1. change the *option* text, without changing the value
2. change the *order* of options, even when options are the same

The first one is necessary for this ticket - where the concept of
'other' premise has an updated label (and yet we're implementing this in
a way that preserves the current implementation that's accessible in
live).

The second one will be necessary for future tickets (specifically
destination premises type), where the order for each option has been
user test specifically & helps users answer the question.

### DIFF
--- a/src/server/common/model/answer/origin-type/origin-type.js
+++ b/src/server/common/model/answer/origin-type/origin-type.js
@@ -1,10 +1,23 @@
 import { RadioButtonAnswer } from '../radio-button/radio-button.js'
 /** @import {RadioButtonConfig} from '../radio-button/radio-button.js' */
+/** @import {RawApplicationState} from '~/src/server/common/model/state/state-manager.js' */
 
 /**
  * export @typedef {'tb-restricted-farm' | 'afu' | 'other'} OriginTypeData
  * @typedef {{ originType: OriginTypeData }} OriginTypePayload
  */
+
+/**
+ * @param {RawApplicationState} app
+ * @returns boolean
+ */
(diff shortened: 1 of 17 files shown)


src/server/common/model/answer/origin-type/origin-type.js
```diff
@@ -1,26 +1,53 @@
...
+  },
+  other: { label: 'Another origin' }
+}
+
 /** @augments {RadioButtonAnswer<OriginTypePayload>} */
 export class OriginTypeAnswer extends RadioButtonAnswer {
-  /** @type {RadioButtonConfig} */
-  static config = {
+  /** @type {RadioButtonConfigFactory} */
+  static config = (app) => ({
```
DrogoNevets, line 46: I see what you're trying to do
but I disagree that this should be a static function, it is using the state, effectively of the application, and thus should be done within the answer, which would mean it could use `this.value` rather than any complications to pass context in

DrogoNevets: as typified by adding the `new` keyword to the tests

DrogoNevets: even if you need to pass app state in to base result on other sections, i feel it should still be instatiated and not static

DrogoNevets (changes requested):

DrogoNevets (approved):

PR #236, opened 2025-02-20
DSFAAP-859: add 'off the farm' origin type options when biosecurity flag is enabled 

hughfdjackson:
merge https://github.com/DEFRA/apha-apps-perms-move-animal-ui/pull/234 & rebase before writing tests for this


src/server/common/model/answer/origin-type/origin-type.js
```diff
@@ -19,12 +20,28 @@ const afuOption = {
   hint: 'Including enhanced with grazing (AFUE)'
 }
 
-const offFarmOptions = {
+const offFarmOptions = () =>
```
DrogoNevets, line 23: if moved down to below the declarations, this wouldnt need to be a function meaning the ternary on line 65 would look similar for both conditions (ie neither is a function call) - this would maintain the same composition and reduce call stack

DrogoNevets (approved):

src/server/common/model/answer/origin-type/origin-type.js
hughfdjackson: @DrogoNevets  - it's in here so that it's evaluated late so we can test it with `spyOnConfig`

PR #243, opened 2025-02-21
DSFAAP-870: add destination type options for movements on to the farm

hughfdjackson:


src/server/common/model/answer/destination-type/destination-type.js
```diff
@@ -1,25 +1,34 @@
...
   hint: 'Including enhanced with grazing (AFUE)'
 }
 
-const onFarmOptions = {
-  'tb-restricted-farm': tbRestrictedOption,
-  afu: afuOption,
-  zoo: { label: 'Zoo' },
-  lab: { label: 'Laboratory' },
-  other: { label: 'Another destination' }
+const onFarmOptions = (app) => {
```
hughfdjackson, line 18: Can be unnested to: 

```javascript
const onFarmOptions = (app) => isOriginAfu ? { ... } : { ... } 
```

PR #249, opened 2025-02-25
[848] PDF compression

DrogoNevets:


.devrc
```diff
@@ -0,0 +1 @@
+export GS_BINARY=/opt/homebrew/bin/gs
```
joseluisgraa, line 1: Do we all need this file? If not maybe add it to gitignore

joseluisgraa: I'm using an envrc file currently

PR #252, opened 2025-02-25
Modifications

eoin-corr-git:


user-journey-tests/page-objects/origin/newAddressPage.js
```diff
@@ -181,6 +181,7 @@ class NewAddressPage extends Page {
     }
 
     await super.selectContinue()
+    if (nextPage) await page.waitForPagePath(nextPage.pagePath)
```
hughfdjackson, line 184: Is there ever a case where we don't want to wait?  If so, is this just a temporary if statement (to disappear once everything is moved over to the new way of doing things)?

hughfdjackson (approved):

PR #265, opened 2025-02-26
Quantity options page

joseluisgraa:


src/server/common/model/answer/quantity-options/quantity-options.js
```diff
@@ -0,0 +1,27 @@
...
+ * @typedef {{ quantityOptions: QuantityOptionsData }} QuantityOptionsPayload
+ */
+
+/**
+ * @augments {RadioButtonAnswer<QuantityOptionsPayload>}
+ */
+export class QuantityOptionsAnswer extends RadioButtonAnswer {
+  /** @type {RadioButtonConfig} */
+  static config = {
+    payloadKey: 'quantityOptions',
```
hughfdjackson, line 16: I'd rename this key as "movingMoreThan75Animals", because otherwise the raw data won't make sense.

src/server/common/model/answer/text-area/text-area.js
```diff
@@ -10,19 +10,20 @@ import { NotImplementedError } from '../../../helpers/not-implemented-error.js'
  * @returns {Joi.Schema}
  */
 const textAreaSchema = ({ payloadKey, validation }) => {
-  let stringValidation = Joi.string().trim().max(validation.maxLength.value)
+  let stringValidation = Joi.string()
+    .required()
+    .trim()
+    .max(validation.maxLength.value)
```
hughfdjackson, line 16: Would be preferable to be a separate PR, but this looks like a really good change :+1:

src/server/destination/quantity-options/index.test.js
```diff
@@ -0,0 +1,67 @@
...
+const questionKey = 'quantityOptions'
+const view = 'common/model/page/question-page.njk'
+const pageUrl = '/destination/quantity-options'
+
+describe('QuantityOptionsPage', () => {
+  let page
+
+  beforeEach(() => {
+    page = new QuantityOptionsPage()
+  })
```
hughfdjackson, line 18: :D same comment as the previous here - 

```
const page = new QuantityOptionsPage()
```

preferred

src/server/common/model/answer/quantity-options/quantity-options.js
hughfdjackson: Actually, it's the questionKey that really matters, rather than this one.

hughfdjackson (approved):

PR #291, opened 2025-03-03
[922] not support for movement type

DrogoNevets:
- **logic added**
- **new exit page**


src/server/common/test-helpers/journey-state.js
```diff
@@ -101,7 +101,7 @@ export const applicationStateWithAnimalIdentifiersSection = {
   origin: {
```
hughfdjackson, line 101: This variable probably wants renaming from `applicationStateWithAnimalIdentifiersSection`, (and 'identification' removed), since that section is no longer part of the application state. 

Should have caught this in a previous PR ! but only noticed it now.

src/server/task-list/controller.test.js
```diff
@@ -72,6 +72,17 @@ describe('#taskListController', () => {
   })
 
   it('Should return the correct task list items when there is state enough to show all sections', async () => {
+    const { origin, destination } = applicationStateWithAnimalIdentifiersSection
+    await session.setState('origin', {
```
hughfdjackson, line 76: I think these overrides can be removed now?  because they match what is currently in `applicationStateWithAnimalIdentifiersSection`

hughfdjackson (approved):

src/server/task-list/controller.test.js
```diff
@@ -72,6 +72,17 @@ describe('#taskListController', () => {
...
+      ...origin,
+      onOffFarm: 'on',
+      originType: 'market'
+    })
+    await session.setState('destination', {
+      ...destination,
+      destinationType: 'tb-restircted-farm'
```
joseluisgraa, line 83: typo here

src/server/task-list/controller.test.js
```diff
@@ -72,6 +72,17 @@ describe('#taskListController', () => {
   })
 
   it('Should return the correct task list items when there is state enough to show all sections', async () => {
```
joseluisgraa, line 74: are we removing the test that covers the scenario where all the sections are visible? (that's what the test description says so I guess we need another test or worst case we need to rename this one)

src/server/task-list/controller.test.js
DrogoNevets: done

PR #304, opened 2025-03-10
[937] new email layout

DrogoNevets:


src/server/check-answers/__snapshots__/index.test.js.snap
```diff
@@ -548,6 +611,11 @@ some disinfectant
 15
 ## How will you reduce the risk of infection from badgers and wildlife?
 some text
+# Biosecurity map
+
+---
 ## Upload a biosecurity map
+
+## undefined
```
hughfdjackson, line 619: We seem to have picked up an `undefined` here that we wouldn't want?

src/server/check-answers/index.js
```diff
@@ -129,16 +129,22 @@ export class SubmitPageController extends QuestionPageController {
...
+    /**
+     * @type {string[]}
+     */
+    const lines = []
+
+    Object.values(application.tasks).forEach((task) => {
+      lines.push(`# ${task.config.title}`)
+      lines.push('')
+      lines.push('---')
+      task.questionPageAnswers.forEach(({ page, answer }) => {
```
hughfdjackson: We've no longer filtering for interstitial.  I'm trying to figure out whether we did that for a non-type reason in the first place :thinking:

PR #307, opened 2025-03-10
[941] make changes required by policy meeting

DrogoNevets:


src/server/biosecurity/kept-separately/index.js
```diff
@@ -19,7 +19,7 @@ export class KeptSeparatelyPage extends QuestionPage {
     if (answer.value === 'yes') {
       return grazingPage
     }
-    return peopleDisinfectionPage
+    return manureAndSlurryPage
```
hughfdjackson, line 22: Needs to be the manure and slurry details page (which we don't have yet).  Typically we've been implementing a stub page (that isn't connected to the routes) that has the right url path & is only a `Page` (not a `QuestionPage`)

hughfdjackson (approved):

PR #309, opened 2025-03-11
[943] how seperated page updates

DrogoNevets:
- **updated page**
- **added missing file**

### DIFF
--- a/src/server/biosecurity/grazing-field-how-separated/index.njk
+++ b/src/server/biosecurity/grazing-field-how-separated/index.njk
@@ -1,0 +1,21 @@
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% extends 'layouts/questions.njk' %}
+{% block heading %}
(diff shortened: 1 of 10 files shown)


src/server/biosecurity/grazing-field-how-separated/index.njk
```diff
@@ -0,0 +1,21 @@
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% extends 'layouts/questions.njk' %}
+{% block heading %}
+  <h1 class="govuk-heading-l">Separating the incoming cattle</h1>
```
hughfdjackson, line 4: weirdly, it doesn't seem to have  made any visual difference, but we've been wrapping these elsewhere in:

```html
{% block heading %}
  <div class="govuk-grid-row">
    <div class="govuk-grid-column-two-thirds">
      <h1 class="govuk-heading-l">{{ heading }}</h1>
    </div>
  </div>
{% endblock %}
```

hughfdjackson (approved):

PR #325, opened 2025-03-14
PDF upload limits and refactor

joseluisgraa:


src/server/common/helpers/file/size.js
```diff
@@ -0,0 +1,20 @@
...
+ * @returns {number}
+ */
+export const fileSizeInMB = (sizeInBytes) => sizeInBytes / (1024 * 1024)
+
+/**
+ * @param {number} sizeInMegaBytes - The file size in megabytes.
+ * @returns {number} The file size in bytes.
+ */
+export const fileSizeInBytes = (sizeInMegaBytes) =>
+  sizeInMegaBytes * 1024 * 1024
```
hughfdjackson, line 12: Went on a bit of a dig about this, because [MB could either be decimal (1000 ^ 2) or binary (1024 ^ 2)](https://en.wikipedia.org/wiki/Megabyte).

In the v2 notify service, this is where [email ends up](https://github.com/alphagov/notifications-api/blob/main/app/v2/notifications/post_notifications.py#L111), which ultimately leads to [`process_sms_or_email_notification`](https://github.com/alphagov/notifications-api/blob/main/app/v2/notifications/post_notifications.py#L162).

From there you get to the [actual file upload](https://github.com/alphagov/notifications-api/blob/main/app/v2/notifications/post_notifications.py#L273).  That's handled by another service - the [document-download-api](https://github.com/alphagov/document-download-api/blob/main/app/upload/views.py#L67), which compares the file to a [MAX_DECODED_FILE_SIZE](https://github.com/alphagov/document-download-api/blob/e8e5cb6396c5c3eb6af3e4b3ad5b9e79e0f662e7/app/upload/views.py#L34).

And [that defaults to](https://github.com/alphagov/document-download-api/blob/e8e5cb6396c5c3eb6af3e4b3ad5b9e79e0f662e7/app/config.py#L41):

src/server/check-answers/index.js
```diff
@@ -105,7 +105,16 @@ export class SubmitPageController extends QuestionPageController {
...
       ) {
-        const compressedFile = await this.handleBiosecurityFile(req)
+        const compressedFile = await handleUploadedFile(
+          req,
+          req.yar.get('biosecurity-map')['upload-plan'],
+          this.logger
+        )
+
+        // Error if the file after compression is still too large
+        if (fileSizeInMB(compressedFile.length) > 2) {
```
hughfdjackson, line 115: You're doing the comparison at exactly the right place - before base64 encoding - as per the other comment.

src/server/common/helpers/file/file-utils.js
```diff
@@ -1,18 +1,25 @@
...
+import { GetObjectCommand } from '@aws-sdk/client-s3'
+import { config } from '~/src/config/config.js'
 
 /**
  * Handles the uploaded file by processing it based on its content type.
  * Compresses the file if it is a PDF or an image, logs the compression details,
  * and returns the processed file.
  * @param {object} req - The request object containing the S3 method to retrieve the file.
- * @param {string} filePath - The path to the file in the S3 bucket.
+ * @param {object} uploadedFile - The path to the file in the S3 bucket.
```
hughfdjackson, line 12: Does the description here need updating?

hughfdjackson: Also, It would be ideal if we can be more specific here & in logger than `{object}` - is that something you've tried & run into some issues with?

src/server/check-answers/index.test.js
```diff
@@ -375,6 +340,69 @@ describe('#CheckAnswers', () => {
     expect(statusCode).toBe(statusCodes.redirect)
     expect(content).toMatchSnapshot('email-content-biosec-enabled')
   })
+
+  it('should compress the file and send email in correct format for flag enabled', async () => {
```
hughfdjackson, line 344: bit of a nit-pick, but it should *handle* the file. It 's only getting compressed if it's over 2MB

src/server/check-answers/index.test.js
```diff
@@ -375,6 +340,69 @@ describe('#CheckAnswers', () => {
...
+    spyOnConfig('featureFlags', { biosecurity: true })
+
+    await session.setState(
+      'origin',
+      validApplicationStateWithBioSecurity.origin
+    )
+    await session.setState(
+      'destination',
+      validApplicationStateWithBioSecurity.destination
+    )
```
hughfdjackson, line 354: Aren't these already set in the `beforeEach` block at the top of the file?

src/server/check-answers/index.test.js
```diff
@@ -375,6 +340,69 @@ describe('#CheckAnswers', () => {
...
```
hughfdjackson, line 385: The same comment here as above:

> Aren't these already set in the beforeEach block at the top of the file?

src/server/check-answers/index.js
hughfdjackson: yes - it's just me re-assuring myself that our checks align with the GDS notify checks.

PR #340, opened 2025-03-25
App model updates, UI working, unit tests broken

hughfdjackson:


src/server/biosecurity-map/upload-progress/index.test.js
```diff
@@ -34,18 +34,22 @@ describe('#UploadPlan', () => {
 
   beforeEach(async () => {
     session = await SessionTestHelper.create(server)
-    await session.setState(uploadConfig.questionKey, {
-      statusUrl: 'http://localhost/status'
+    await session.setState('application', {
```
hughfdjackson, line 37: I don't honestly know how this ever worked, looking at the code.  I think this `await` statement might be dead code.

src/server/check-answers/index.test.js
```diff
@@ -27,10 +27,8 @@ const {
   origin,
   destination,
   licence,
-  identification,
-  biosecurity,
   'biosecurity-map': biosecurityMap
-} = validApplicationState
+} = validApplicationState.application
```
hughfdjackson, line 31: We don't need this additional nesting - the validApplicationState **is** what's stored under `req.yar.get('application')`

src/server/common/model/answer/radio-button/radio-button.js
```diff
@@ -34,7 +34,9 @@ const createRadioSchema = (config) => {
  */
 const handleConfig = (context, config) => {
   if (typeof config === 'function') {
-    return config(context ?? {})
+    return config(
+      context ?? /** @type {RawApplicationState} */ ({ application: {} })
```
hughfdjackson, line 38: `RawApplicationState` shouldn't have `application` nested within it.  It should look like it looked before now, e.g.:

```javascript
{
  origin: { ... }
}
```

src/server/common/model/state/state-manager.js
```diff
@@ -4,9 +4,11 @@
 
 /**
  * @typedef {Record<string, any>} RawSectionState
- * @typedef {Record<string, RawSectionState>} RawApplicationState
+ * @typedef {{ application: RawSectionState}} RawApplicationState
+ * @typedef  {'origin' |  'destination' | 'licence' | 'identification' | 'biosecurity' | 'biosecurity-map'} SectionKeys
```
hughfdjackson, line 8: As per our conversation elsewhere, I think this is where the complexity in the refactor has creeped in. 

Originally, RawApplicationState looks like (for instance)

```javascript
{
  origin: { 
    onOffFarm: 'on',
    address: { ... }
  },
  destination: { ... }
}
```

If we keep this interface the same, then the refactoring surface gets a lot smaller.  (Also, I think the structure is neater, because we avoid one level of nesting which isn't serving a strong purpose)

PR #341, opened 2025-03-25
Initial changes

joseluisgraa:


src/server/common/model/state/state-manager.js
```diff
@@ -24,21 +24,27 @@ export class StateManager {
...
-    return Object.fromEntries(
-      sectionKeys
-        .map((key) => [key, this._request.yar.get(key)])
-        .filter((entry) => entry.at(1) !== null)
+    return (
+      Object.fromEntries(
+        sectionKeys
+          .map((key) => [key, this._request.yar.get('application')?.[key]])
+          .filter((entry) => entry.at(1) !== null)
+      ) || {}
```
hughfdjackson, line 32: I think we can just return `this._request.yar.get('application') ?? {}`?

hughfdjackson: and remove `sectionKeys` as a result.

src/server/common/model/state/state-manager.test.js
```diff
@@ -54,43 +54,29 @@ describe('StateManager.toState', () => {
     const partialState = { origin }
     const request = testRequest(partialState)
     const state = new StateManager(request)
-    expect(state.toState()).toEqual(partialState)
+    expect(state.toState()).toStrictEqual(partialState)
```
hughfdjackson, line 57: I think we want to also test that request.yar.get was called with `application`?

src/server/common/model/state/state-manager.js
joseluisgraa: done

src/server/common/model/state/state-manager.test.js
joseluisgraa: done

hughfdjackson (approved):

PR #355, opened 2025-04-03
Updated the logic to calculate next page after destination farm address

joseluisgraa:


src/server/common/model/answer/destination-type/destination-type.test.js
```diff
@@ -30,62 +30,96 @@ describe('DestinationType', () => {
       'Select where the animals are going'
     )
   })
-})
 
-describe('DestinationType.config.options', () => {
```
hughfdjackson, line 35: Feels like this indentation change is neither fish nor fowl.  The two conventions we have: 

```javascript
describe('ClassName', () {
  describe('method', () => { ... })
})
```

Or

```javascript
describe('ClassName.method', () => {})
describe('ClassName.otherMethod', () => {})
```

This is now in between those two options?


(FWIW, I prefer the second convention - easier to read with less nesting in most cases)

PR #366, opened 2025-04-08
[1050] Added Animal Identification section

DrogoNevets:


user-journey-tests/specs/identification/taskListVerification.spec.js
```diff
@@ -0,0 +1,35 @@
...
+import signInPage from '../../page-objects/signInPage.js'
+
+describe('Identification - task list appearance', () => {
+  // eslint-disable-next-line no-undef
+  before(async () => {
+    await loginAndSaveSession(signInPage)
+    await taskListPage.navigateToPageAndVerifyTitle()
+  })
+
+  beforeEach(async () => [await restoreSession()])
```
hughfdjackson, line 16: square brackets here should be curlies: 

```javascript
beforeEach(async () => { await restoreSession() })
```

To be fair, even that is redundant - the `beforeEach` function is expecting a function that takes no arguments & returns a promise (aka a nullary async function) - which is what restoreSession already is.

So we can do:

```javascript
beforeEach(restoreSession)
```

user-journey-tests/specs/identification/taskListVerification.spec.js
```diff
@@ -0,0 +1,35 @@
...
+    await enableIdentification()
+    await taskListPage.verifyStatus({
+      position: 4,
+      taskTitle: 'Animal identification',
+      expectedStatus: 'Incomplete'
+    })
+  })
+
+  it('Should verify identification appears when zoo options selected', async () => {
+    await enableIdentification(true, true)
```
hughfdjackson, line 28: More readable to go for an options object here: 

```javascript
await enableIdentification({
  originZoo: false,
  destinationZoo: false
})
```

user-journey-tests/helpers/testHelpers/flow-helpers/enableIdentification.js
```diff
@@ -0,0 +1,68 @@
...
+    },
+    howManyAnimalsPage
+  )
+  await howManyAnimalsPage.inputTextAndContinue('550', reasonForMovementPage)
+  await reasonForMovementPage.selectOtherAndContinue(additionalInfoPage)
+  await additionalInfoPage.selectContinue()
+  await waitForPagePath(destinationAnswersPage.pagePath)
+  await destinationAnswersPage.selectContinue()
+  await taskListPage.verifyPageHeadingAndTitle()
+}
```
hughfdjackson, line 68: Did you have a look at the existing origin & destination helpers, and conclude that having this separate helper was easier to maintain overall?

hughfdjackson (changes requested):

user-journey-tests/helpers/testHelpers/flow-helpers/enableIdentification.js
hughfdjackson: I think I'm inclined to agree that this is the best option we have for the moment.

PR #375, opened 2025-04-11
[1056] calf ear tag page

DrogoNevets:


src/server/identification/ear-tags-calves/index.js
```diff
@@ -0,0 +1,39 @@
...
+const customHeading = 'Official ear tag numbers for calves under 42 days old'
+
+export class EarTagsCalvesPage extends QuestionPage {
+  urlPath = '/identification/enter-ear-tags-calves'
+  sectionKey = 'identification'
+  question =
+    'Enter the ear tag numbers of the calves under 42 days old you are planning to move'
+
+  questionKey = 'earTagsCalves'
+  Answer = EarTagsAnswer
```
joseluisgraa, line 15: I think the errors are wrong because you are reusing EarTagsAnswer here and I guess it has to be a different model with its own error messages

PR #377, opened 2025-04-14
[1048] Updated homepage copy

DrogoNevets:


eoin-corr-git:
Tests didn't rerun for some reason but passing locally so all good to merge

PR #383, opened 2025-04-16
DSFAAP-1056: suggested test improvements & refactor

hughfdjackson:
We want our date comparisons to:

- always represent the date submitted by the user as though it's a Europe/London date (e.g. accounting for BST in the summer months)
- no matter what timezone the system time is set to, use the date it would be in `Europe/London` at that point in time as our point of comparison

 I think these tests help to capture that coverage more accurately - and they've shown that a simpler implementation can have the desired effect.

### DIFF
--- a/src/server/common/model/answer/date/date-utils.test.js
+++ b/src/server/common/model/answer/date/date-utils.test.js
@@ -4,7 +4,6 @@ import {
   isFutureDate,
   toBSTDate
 } from './date-utils.js'
-import { subDays } from 'date-fns'
 
 describe('Date Helpers', () => {
   describe('toBSTDate', () => {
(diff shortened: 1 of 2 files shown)


src/server/common/model/answer/date/date-utils.test.js
```diff
@@ -28,41 +27,52 @@ describe('Date Helpers', () => {
   })
 
   describe('isFutureDate', () => {
+    beforeEach(() => {
+      jest.useFakeTimers()
+      jest.setSystemTime(new TZDate('2025-04-01T00:00:00', 'Europe/London'))
```
joseluisgraa, line 32: If we are only using TZDate in tests you might want to move that package to devDependencies

hughfdjackson: We still have one usage in `toBSTDate` in the implementation - great thing to check though :+1:

PR #388, opened 2025-04-17
[FIXUP] [630] use the POST - REDIRECT - GET pattern

DrogoNevets:


src/server/common/controller/question-page-controller/question-page-controller.test.js
```diff
@@ -593,28 +582,21 @@ describe('QuestionPageController', () => {
...
-
-      const viewArgs = h.view.mock.calls[0][1]
-      expect(viewArgs.answer).toBeInstanceOf(TestAnswer)
-
-      expect(result).toBe('view')
+          errors: {
+            [questionKey]: { text: 'There is a problem' }
+          }
+        }
+      ])
```
hughfdjackson, line 599: Aren't we also expecting this to persist the submitted payload (for replay during calls to the getHandler)?

src/server/common/controller/question-page-controller/question-page-controller.test.js
```diff
@@ -327,31 +327,24 @@ describe('QuestionPageController', () => {
...
-        expect(payload).toEqual(expect.stringContaining('There is a problem'))
-        expect(
-          /** @type {HTMLInputElement} */ (
-            document.querySelector(questionElementSelector)
-          )?.value
-        ).toBe('')
-
-        expect(statusCode).toBe(statusCodes.ok)
+        expect(statusCode).toBe(statusCodes.redirect)
+        expect(headers.location).toBe(questionUrl)
```
hughfdjackson, line 338: It feels like we need one extra case - covering the page being loaded by a user after this redirect. 

e.g. given there is an error record in the state, the GET page renders the error view as expected?

src/server/common/controller/question-page-controller/question-page-controller.test.js
hughfdjackson: Also, 'toStrictEqual' is a little stronger of a matcher - see https://jestjs.io/docs/using-matchers

src/server/common/controller/question-page-controller/question-page-controller.js
```diff
@@ -52,19 +53,54 @@ export class QuestionPageController extends GenericPageController {
...
+          ? this.page.Answer.fromState(pageError?.value)
+          : answer,
+        ...errorProps,
+        viewModelOptions: Object.assign(
+          {
+            validate: false,
+            question: this.page.question
+          },
+          errorProps.viewModelOptions
+        ),
```
hughfdjackson, line 81:
```
viewModelOptions: { 
  validate: false,
  question: this.page.question,
  ...errorProps.viewModelOptions
}
```
?

DrogoNevets: the rules cdp have put in don't like that, as it means a property is overridden (they want the spread first - but that breaks what we are trying to do

so actually we can get rid of the spread altogether now it is in its own seperate branch 

done

src/server/common/controller/question-page-controller/question-page-controller.js
```diff
@@ -52,19 +53,54 @@ export class QuestionPageController extends GenericPageController {
...
+      errorProps.viewModelOptions = {
+        validate: true,
+        question: this.page.question
+      }
+
+      return h.view(this.page.view, {
+        nextPage: req.query.redirect_uri,
+        heading: this.page.heading,
+        answer: pageError
+          ? this.page.Answer.fromState(pageError?.value)
```
hughfdjackson, line 72: We've discussed all of this in slack, but just to make sure there's a 'to fix' tick list here: 

- I believe using `fromState` here creates the potential for bugs, since `.value` (called in the postHandler) alters the data subtly (e.g trimming or substituting in empty strings).  We're missing the test coverage to prove the previous approach & this one are equivalent, so it's probably easier to store the payload in the error state, then use `new this.page.Answer(pageError.payload, applicationState)`
- we're missing `applicationState` here, which causes the radio button options (e.g.) not to be filtered correctly when displaying an error

hughfdjackson (approved):

PR #391, opened 2025-04-23
POST-REDIRECT-GET query params

DrogoNevets:


src/server/common/controller/question-page-controller/question-page-controller.js
```diff
@@ -109,7 +110,7 @@ export class QuestionPageController extends GenericPageController {
         payload
       })
 
-      return h.redirect(this.page.urlPath)
+      return h.redirect(nextPageRedirect(this.page, req.query))
```
hughfdjackson, line 113: I think we're missing a test case for this in `question-page-controller.test.js`?  

We've got great coverage of the logic (in the extracted nextPageRedirect), but not the glue (that we're calling that function, that we pass the right things in)

PR #392, opened 2025-04-23
Removed proxyFetch and replaced with Wreck

joseluisgraa:


src/server/common/connectors/notify/notify.integration.test.js
```diff
@@ -13,7 +13,7 @@ describe('sendNotification (integration)', () => {
     const configGet = config.get.bind(config)
     const notifyConfig = {
       ...config.get('notify'),
-      timeout: 0
+      timeout: 1 // setting timeout to 1ms as 0ms means no timeout when using Wreck
```
hughfdjackson, line 16: this looks liable to be flakey unless we introduce an explicit await in the test that definitely exceeds 1ms?

src/server/common/connectors/notify/notify.js
```diff
@@ -27,31 +28,28 @@ export async function sendNotification(data) {
...
+      payload,
       headers: {
         Authorization: 'Bearer ' + createToken(notifyConfig.apiKey)
       },
-      signal: AbortSignal.timeout(timeout)
+      timeout
     })
   } catch (err) {
-    if (err.code && err.code === err.TIMEOUT_ERR) {
+    if (err.output?.statusCode === statusCodes.gatewayTimeout) {
```
hughfdjackson, line 39: It's a bit suspect to me that this works, if I'm 100% honest.  A gateway timeout is delivered to the client from a server that's acting as a proxy: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/504

it's not what I'd expect to see from a HTTP client that has chosen to time _itself_ out. :thinking:

hughfdjackson: And yet this is clearly [what Wreck is doing](https://github.com/hapijs/wreck/blob/ba28b0420d6b0998cd8e61be7f3f8822129c88fe/lib/index.js#L309)...
image

src/server/common/connectors/notify/notify.js
```diff
@@ -27,31 +28,28 @@ export async function sendNotification(data) {
...
       throw new Error(`Request to GOV.uk notify timed out after ${timeout}ms`)
+    } else if (err.data) {
+      const errors = JSON.parse(err.data.payload?.toString())
```
hughfdjackson, line 42: I think this is good as is - in a 'let it crash' sort of way - but if payload is undefined, then so will `payload?.toString()` be. 

That'll cause a JSON.parse syntax error to be thrown.  In this case, I think that's what we want.  It'll be easy to get to the stack trace back to here and see what we want, and the user will see the outcome we intended

src/server/common/connectors/notify/notify.js
```diff
@@ -27,31 +28,28 @@ export async function sendNotification(data) {
...
+      const errorMessages = errors.errors.map((error) => error.message)
```
hughfdjackson, line 43: Similar comment here :+1:

src/server/common/connectors/notify/notify.test.js
```diff
@@ -12,6 +17,12 @@ const testData = {
   }
 }
 
+const timeout = config.get('notify').timeout
+const templateId = config.get('notify').templateId
+const caseDeliveryEmailAddress = config.get('notify').caseDeliveryEmailAddress
+
+jest.mock('@hapi/wreck')
```
hughfdjackson, line 24: Since we spy on Wreck.post, I believe this isn't needed?

hughfdjackson (changes requested):
Just one change actually requested - adding a 2-5ms timeout to the notify integration test.

src/server/common/connectors/notify/notify.integration.test.js
hughfdjackson: Ah - I'm totally wrong here.  we're awaiting the eventual result of the http call

hughfdjackson (approved):

hughfdjackson (approved):

PR #401, opened 2025-04-28
Escaping HTML lt and gt

joseluisgraa:


src/server/common/helpers/html-escape.js
```diff
@@ -0,0 +1,3 @@
+export const escapeHtml = (unsafe) => {
+  return unsafe?.replaceAll('<', '&lt;').replaceAll('>', '&gt;')
```
hughfdjackson, line 2: The only outstanding question I have is whether this is sufficient to prevent all forms of HTML injection attack. 

This OWASP guide suggests that we should go further: https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.md#output-encoding

hughfdjackson (commented):
Mind to have a read of the [OWASP recommendations](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.md#output-encoding) & see if we're escaping all of the values we ought to be?

PR #408, opened 2025-04-30
Dsfaap 1108 confirmation email

joseluisgraa:


src/config/config.js
```diff
@@ -284,6 +272,24 @@ export const config = convict({
       format: Boolean,
       default: false,
       env: 'NOTIFY_EMAIL_DOWNLOAD_CONFIRMATION'
+    },
+    caseDeliveryTemplateId: /** @type {SchemaObj<string | null>} */ {
```
DrogoNevets, line 276: would it not make sense to have a `caseDeliveryObject` with the 2 props?

DrogoNevets (changes requested):

src/config/config.js
joseluisgraa: I was expecting that comment tbh, I wasn't sure whether it was worth it or not, but I'll create two, one for caseDelivery and one for applicant

joseluisgraa: done

DrogoNevets (approved):

PR #422, opened 2025-05-08
[1123] update origin labels

DrogoNevets:
- **add new page and tests**
- **updated lab and other options**


hughfdjackson:
@DrogoNevets - there's another cleanup that this enables - we have a list of 'possibly TB restricted' origins, but we can use `OriginTypeAnswer.isTbRestricted`?

PR #423, opened 2025-05-08
Removed animal identifiers toggle

joseluisgraa:


src/server/destination/destination-type/index.js
```diff
@@ -38,14 +36,7 @@ export class DestinationTypePage extends QuestionPage {
...
   nextPage(answer, context) {
-    if (
-      !config.get('featureFlags').animalIdentifiers &&
-      context.origin?.onOffFarm === 'on' &&
-      restricted.includes(context.origin?.originType) &&
-      possibleRestricted.includes(answer.value)
-    ) {
-      return destinationNotSupportedPage
-    } else if (context.origin?.onOffFarm === 'on') {
+    if (context.origin?.onOffFarm === 'on') {
```
hughfdjackson, line 39: Do we use the 'possibleRestricted' and 'restricted' arrays any more?  I think the can be removed / moved to the test file at least.

PR #426, opened 2025-05-12
Dsfaap 1126 bio sec interstitial page

joseluisgraa:


src/server/common/model/answer/hidden/hidden.js
```diff
@@ -0,0 +1,91 @@
...
+
+  /**
+   * @returns {TextData}
+   */
+  toState() {
+    return this.value ?? ''
+  }
+
+  validate() {
+    return { isValid: !!this.value, errors: {} }
```
hughfdjackson, line 54: `this.value === this.config.value`?  I think technically they're the same thing - so honestly this is probably just that this feels a bit indirect to me.

joseluisgraa: This is just checking that the value is set to something (it is truthy)

joseluisgraa: But might be safer to check it is the same as the configured value. I'll update it.

hughfdjackson (approved):

PR #436, opened 2025-05-29
[1193]  send application to CM API microservice

DrogoNevets:
- **DSFAAP-1103: WIP - construct basic payload for submission to case management api**
- **send application to case management api, and get reference**


DrogoNevets:
Super rough draft, allows a happy path of sending a simple qpplication through to the CM API microservice

eoin-corr-git (approved):

PR #440, opened 2025-06-02
[1193] add querystring flag for using CMAPI

DrogoNevets:


src/server/check-answers/index.js
```diff
@@ -93,6 +93,49 @@ export class SubmitPageController extends QuestionPageController {
...
+
+    const application = ApplicationModel.fromState(applicationState)
+
+    const { isValid: isValidApplication } = application.validate()
+
+    if (isValidPage && isValidApplication) {
+      if (
+        config.get('featureFlags').sendToCaseManagement ||
+        (req.query['case-management-api'] === 'true' &&
+          !config.get('featureFlags').sendToCaseManagement)
```
hughfdjackson, line 123: I think this line is redundant.  On this side of the `or` expression, we know that `config.get('featureFlags').sendToCaseManagement` is at least falsey already.

PR #455, opened 2025-06-17
origin other page implemented

DrogoNevets:


src/server/common/model/answer/origin-type-other/origin-type-other.js
```diff
@@ -0,0 +1,36 @@
...
+      maxLength: {
+        value: 5000,
+        message: 'Your answer must be no longer than 5000 characters'
+      }
+    }
+  }
+
+  get emailHtml() {
+    const data = this._data?.[this.config.payloadKey]
+    if (data.length === 0) {
```
hughfdjackson, line 31: We can't have a zero-length case co-exist with a provided `validation.empty.message` in the config - so we can remove this. 

we should just be passing through the escapedMarkdown of the text answer

src/server/common/model/answer/origin-type-other/origin-type-other.test.js
```diff
@@ -0,0 +1,53 @@
...
+
+const maxLength = 5000
+
+/** @type {OriginTypeOtherPayload} */
+const payload = {
+  originTypeOther: 'other origin type'
+}
+
+describe('OriginTypeOther', () => {
+  it('should be a text area', () => {
```
hughfdjackson, line 13: text area -> text answer

PR #464, opened 2025-06-24
Updated origin/destination options

DrogoNevets:


src/server/common/model/answer/destination-type/destination-type.js
```diff
@@ -6,10 +6,6 @@ import { RadioButtonAnswer } from '../radio-button/radio-button.js'
  * @typedef {{ destinationType: DestinationTypeData }} DestinationTypePayload
  */
 
-const isOnToTheFarm = (app) => app.origin?.onOffFarm === 'on'
-const isOriginAfu = (app) => app.origin?.originType === 'afu'
```
joseluisgraa, line 10: I would still keep these two variables instead of checking the value each time

src/server/destination/destination-type/__snapshots__/index.test.js.snap
```diff
@@ -1,84 +1,6 @@
...
-
-            <button type="submit" class="govuk-button" data-module="govuk-button" id="continue-button">
-  Continue
-</button>
-        </div>
-      </div>
-    </form>
-        "
-`;
+exports[`DestinationTypePage.nextPage off the farm DestinationTypePage.content should render expected response and content 1`] = `undefined`;
```
joseluisgraa, line 3: this doesn't look right

DrogoNevets: in what way?

joseluisgraa (approved):

PR #470, opened 2025-06-27
DSFAAP-1303: making controllers generic

hughfdjackson:


src/server/common/model/section/section-model/section-model.js
```diff
@@ -138,7 +138,7 @@ export class SectionModel {
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   buildGdsTaskDetails(req) {
     const sectionValidity = this.validate()
-    const applicationState = new StateManager(req).toState()
+    const applicationState = new TbStateManager(req).toState()
```
hughfdjackson, line 141: Next step is to parameterise the section model on StateManager also

PR #498, opened 2025-07-03
[1236] numbe rof animals page

DrogoNevets:


src/server/exotics/about/number-of-animals/index.test.js
```diff
@@ -0,0 +1,80 @@
...
+
+const sectionKey = 'about'
+const questionKey = 'numberOfAnimals'
+const pageUrl =
+  '/exotics/about-the-movement/what-is-moving/select-animals/quantity'
+const page = numberOfAnimalsPage
+const question = 'How many animals are you planning to move?'
+
+const payload = {
+  [questionKey]: 'some text'
```
joseluisgraa, line 14: is this correct? should be a number I guess?

PR #546, opened 2025-07-09
[1361] What is the CPH number for premises where the visit is happening?

DrogoNevets:


hughfdjackson (dismissed):
Move to the `location-of-movement` section, otherwise LGTM

PR #559, opened 2025-07-10
AnimalsOnPremisesPage

joseluisgraa:


src/server/common/model/answer/text-area/text-area.js
```diff
@@ -40,7 +40,7 @@ const textAreaSchema = ({ payloadKey, validation }) => {
 /**
  * export @typedef {{
  *  payloadKey: string,
- *  spellcheck?: false,
+ *  spellcheck?: boolean,
```
hughfdjackson, line 43: If you look at line 141 in this file, setting it to anything other than false is a no-op.  i.e. it just skips the setting completely.

That's why we had it narrowed down like this - so when the ticket says false, I've just been missing this off & either:

- letting the snapshot confirm the HTML looks as expected

**or**

- testing `expect(Answer.config.spellcheck).not.toBe(false)`

hughfdjackson (approved):

PR #563, opened 2025-07-10
[1409] remove email flag, and standardise check-answers

DrogoNevets:


src/server/check-answers/index.test.js
```diff
@@ -0,0 +1,296 @@
...
+          {
+            Cookie: session.sessionID
+          }
+        )
+      )
+
+      expect(statusCode).toBe(statusCodes.redirect)
+      expect(headers.location).toBe(confirmationUri)
+
+      const refNum = await session.getRawState('applicationReference')
```
hughfdjackson, line 283: It would be good to namespace this (so there's no data overlap between multiple journeys) - eg. by the StateManager.key

src/server/check-answers/index.js
```diff
@@ -0,0 +1,136 @@
+import { sectionToSummary } from '../common/templates/macros/create-summary.js'
```
hughfdjackson, line 1: We would want to move this all to ~/src/server/common/controllers/..., I expect

src/server/check-answers/index.js
```diff
@@ -0,0 +1,136 @@
...
+
+  view = `check-answers/index`
+
+  Answer = ConfirmationAnswer
+
+  nextPage() {
+    return new ConfirmationPage()
+  }
+
+  viewProps(req) {
```
hughfdjackson, line 45: We can likely make our life easier by moving these view props down into the controller.

I believe super.handlePost takes viewProp args into it.

Then this becomes just a regular question page.

src/server/check-answers/index.js
```diff
@@ -0,0 +1,136 @@
...
+    const application = this.page.ApplicationModel.fromState(applicationState)
+
+    const { isValid: isValidApplication } = application.validate()
+
+    if (isValidPage && isValidApplication) {
+      return await this._handleToCaseManagementApi(req, h)
+    }
+
+    if (!isValidApplication) {
+      return h.redirect('/task-list-incomplete')
```
hughfdjackson, line 127: needs to vary on each journey type

src/server/tb/biosecurity-map/size-error/size-error-page-controller.js
```diff
@@ -1,12 +1,11 @@
-import { submitSummaryPage } from '../../check-answers/index.js'
 import { PageController } from '../../../common/controller/page-controller/page-controller.js'
 import { BiosecurityMapAnswer } from '../../../common/model/answer/biosecurity-map/biosecurity-map.js'
 import { TbStateManager } from '~/src/server/tb/state-manager.js'
 import { uploadPlanPage } from '../upload-plan/index.js'
 
 export class SizeErrorPageController extends PageController {
   nextPage() {
-    return submitSummaryPage
+    return { urlPath: '/tb/submit/check-answers' }
```
joseluisgraa, line 8: is it correct to just define the urlPath and not pass the page as we normally do? just checking...

DrogoNevets: yes including page creates a cyclic dependency

joseluisgraa (approved):

PR #575, opened 2025-07-14
Dsfaap 1405

joseluisgraa:


DrogoNevets (dismissed):

src/server/common/model/page/actionable-exit-page-model.test.js
```diff
@@ -0,0 +1,19 @@
+import { ActionableExitPage } from './actionable-exit-page-model.js'
+
+describe('ActionableExitPage', () => {
+  let page
```
hughfdjackson, line 4: `const page = new ActionableExitPage()` would do the trick :)

DrogoNevets (approved):

PR #584, opened 2025-07-16
Exotics homepage and refactoring of the TB version

joseluisgraa:


src/server/exotics/task-list-incomplete/index.js
```diff
@@ -7,6 +7,8 @@ const pageHeadingAndTitle =
   'You need to complete all of the sections before you review and submit'
 
 export class TaskListIncompletePage extends Page {
+  key = 'taskListIncomplete'
+  sectionKey = 'exotics'
```
hughfdjackson, line 11: Are these only used for namespacing for the purposes of the plugin name? 🤔

src/server/tb/home/index.js
```diff
@@ -0,0 +1,24 @@
...
+
+const pageHeadingAndTitle =
+  'Apply for a Bovine Tuberculosis (TB) movement licence'
+
+export class HomePage extends Page {
+  key = 'home'
+  sectionKey = 'tb'
+  view = 'tb/home/index.njk'
+  urlPath = '/'
+  pageTitle = config.get('serviceName')
```
hughfdjackson, line 15: Title and heading should match, so maybe: 

```
const pageHeadingAndTitle = config.get('serviceName')

...

  pageTitle = pageHeadingAndTitle
  pageHeading = pageHeadingAndTitle

```

Also notice that we've got a property - pageHeading - and a getter `heading` in the base page model, which seems redundant.  Not 100% sure why we've got that 🤔

hughfdjackson (changes requested):
The changes I'd like to see before merging are the ones that causes changes in the TB home page, and also the link that the exotics page links to.

src/server/exotics/task-list-incomplete/index.js
joseluisgraa: yes, if we don't have them then we are registering a plugin with name undefined-undefined and there can only be one, so I had a conflict with your task list incomplete page and had to fix it

templates/content-page/index.js.hbs
```diff
@@ -7,6 +7,8 @@ const pageHeadingAndTitle =
   '{{sentenceCase heading}}'
 
 export class {{pascalCase pageName}}Page extends Page {
+  key = '{{camelCase pageName}}'
+  sectionKey = '{{camelCase sectionKey}}'
```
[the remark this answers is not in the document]
joseluisgraa: I thought I changed that to use the journey, I'll do that now, so they will be namespaced with the journey name when creating the plugin name

src/server/exotics/home/index.njk
```diff
@@ -0,0 +1,65 @@
...
+{%
+  extends
+  'layouts/page.njk'
+%}
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <div class="govuk-body" data-testid="app-page-body">
+        <div class="govuk-!-margin-top-0 govuk-!-margin-bottom-8">
+          <h1 class="govuk-heading-l">Move animals under disease controls</h1>
```
hughfdjackson, line 11: Use `{{ heading }}` as per convention?  Otherwise the `pageHeading` doesn't mean much in the page model

src/server/tb/home/index.js
joseluisgraa: using heading in the template so both are not the same now

src/server/tb/home/index.js
```diff
@@ -0,0 +1,24 @@
+/** @import { ServerRegisterPluginObject } from '@hapi/hapi' */
+
+import { PageController } from '~/src/server/common/controller/page-controller/page-controller.js'
+import { Page } from '~/src/server/common/model/page/page-model.js'
+import { config } from '~/src/config/config.js'
+
+const pageHeadingAndTitle =
+  'Apply for a Bovine Tuberculosis (TB) movement licence'
```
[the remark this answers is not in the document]
joseluisgraa: updated to use heading in the template

PR #590, opened 2025-07-23
[1464] iso unit routing updates

DrogoNevets:


src/server/tb/destination/destination-type/index.js
```diff
@@ -62,6 +63,17 @@ export class DestinationTypePage extends QuestionPage {
       return afuOnlyOffExitPage
     }
 
+    if (
```
joseluisgraa, line 66: might be worth adding to unit tests

joseluisgraa (approved):

eoin-corr-git (approved):

PR #591, opened 2025-07-23
[1465] Routing updates after destination-address for ISO units

DrogoNevets:


src/server/tb/destination/destination-farm-address/index.test.js
```diff
@@ -56,9 +55,20 @@ describe('DestinationFarmAddressPage', () => {
     describe('off the farm', () => {
       const context = { origin: { onOffFarm: 'off' } }
 
-      it('should return summary page', () => {
+      it('should return howManyAnimalsPage if coming from iso0unit', () => {
```
joseluisgraa, line 58: typo

joseluisgraa, line 58: can you add tests for movement on tb restricted and not tb restricted? I think these were missing before.

PR #597, opened 2025-07-24
[1467] New page - what date will the animals move off (ISO units)

DrogoNevets:
- **more routing logic for iso units**
- **added the page**
- **added the page**


src/server/tb/destination/movement-date/__snapshots__/index.test.js.snap
```diff
@@ -0,0 +1,21 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
```
joseluisgraa, line 2: 404

PR #614, opened 2025-07-29
[1505] Update: Which type of movement does your application relate to?

DrogoNevets:


src/server/exotics/about/movement-type/__snapshots__/index.test.js.snap
```diff
@@ -31,7 +31,7 @@ exports[`MovementTypePage content should render expected response and content 1`
       <div class="govuk-radios__item">
         <input class="govuk-radios__input" id="visit" name="movementType" type="radio" value="visit">
         <label class="govuk-label govuk-radios__label" for="visit">
-          A visit on to your farm or premises
+          A short term visit or activity on a farm or premises
```
hughfdjackson, line 34: Sorry about a late update - got feedback from Anita that it should be 'short-term'

hughfdjackson (changes requested):

joseluisgraa (approved):

PR #655, opened 2025-09-01
[1572] [1573] autocomplete list styling

DrogoNevets:


src/client/stylesheets/components/_autocomplete.scss
```diff
@@ -0,0 +1,7 @@
+.autocomplete__wrapper {
+  width: 300px;
```
joseluisgraa, line 2: wondering if we could try to use any of the existing govuk classes to avoid hardcoding a value here (thinking about small screens for example)

joseluisgraa (approved):

PR #668, opened 2025-09-05
Origin section

DrogoNevets:
- **1633 1645 1637**
- **1638**
- **1639**

### DIFF
--- a/src/server/fmd/origin/section.test.js
+++ b/src/server/fmd/origin/section.test.js
@@ -31,11 +31,11 @@ describe('OriginSection', () => {
     expect(OriginSection.config.isVisible({})).toBe(true)
   })
 
-  it('should not be enabled when about section is incompelte', () => {
+  it('should not be enabled when about section is incomplete', () => {
     expect(OriginSection.config.isEnabled({})).toBe(false)
   })
 
-  it('should be enabled when about section is compelte', () => {
+  it('should be enabled when about section is complete', () => {
     expect(
       OriginSection.config.isEnabled({
         about: aboutSectionComplete


src/server/fmd/origin/section.test.js
```diff
@@ -12,12 +13,21 @@ describe('OriginSection', () => {
...
+    expect(
+      OriginSection.firstPageFactory({
+        about: {
+          whatIsMoving: value
+        }
+      })
+    ).toBeInstanceOf(expectedPage)
   })
 
   it('should not be visible', () => {
```
joseluisgraa, line 29: it('should be visible'...

src/server/fmd/origin/premises-type/index.test.js
```diff
@@ -0,0 +1,161 @@
...
+const questionKey = 'premisesType'
+const pageUrl = '/fmd/movement-origin/premises-type'
+const page = premisesTypePage
+const question = 'What type of premises is the origin?'
+
+const payload = {
+  [questionKey]: 'farm'
+}
+
+describe('Answer', () => {
```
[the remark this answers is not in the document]
joseluisgraa: just compare with the tests we add for any other new page we create. We don't need to test stuff that is tested with the snapshot (options, labels...) and we don't test specific data validation in here or html output, we do it in the model.

src/server/fmd/origin/section.test.js
```diff
@@ -12,15 +14,32 @@ describe('OriginSection', () => {
...
-    expect(OriginSection.config.isVisible({})).toBe(false)
+  it('should be visible', () => {
+    expect(OriginSection.config.isVisible({})).toBe(true)
   })
 
   it('should not be enabled', () => {
     expect(OriginSection.config.isEnabled({})).toBe(false)
   })
+
+  it('should be enabled', () => {
```
joseluisgraa, line 38: "should be enabled when the about section is complete", and similar comment for the other test for "not enabled"

PR #707, opened 2025-09-19
[1733] [1741] start page logic, new page

DrogoNevets:


src/server/fmd/licence/section.test.js
```diff
@@ -14,17 +15,30 @@ describe('LicenceSection', () => {
...
   })
 
-  it('should have the correct first page for milk', () => {
+  it('should have the correct first page for milk being moved by prodcuer', () => {
     expect(
       LicenceSection.firstPageFactory({
         about: {
           ...aboutSectionCompleteOtherMovement,
-          whatIsMoving: 'milk'
+          whatIsMoving: 'milk',
```
joseluisgraa, line 23: we have data for this one already in that same test-helpers file: aboutSectionCompleteMilkProducer I think it is called

PR #795, opened 2026-05-22
Updated routing for movement ON when origin is NOT AFU

joseluisgraa:


src/server/tb/destination/destination-type/index.js
```diff
@@ -47,13 +47,26 @@ export class DestinationTypePage extends QuestionPage {
   }
 
   _onFarmNextPage(answer, context) {
-    if (context.origin?.originType === 'afu' && answer.value === 'other') {
-      return afuOnlyOnExitPage
+    // Origin AFU
+    if (context.origin?.originType === 'afu') {
```
mukeshrawat02: Looks good. Just a small non-blocking suggestion: I think to improve readability and avoid repeating `context.origin?.originType` and `answer.value`, we could assign them to local variables and use them throughout.. something like:

```
const originType = context.origin?.originType
const destination = answer.value

if (originType === 'afu') {
  if (destination === 'other') {
    return afuOnlyOnExitPage
  }

  if (destination === 'afu') {
    return destinationFarmCphPage
  }
}

if (destination === 'other') {
  return destinationTypeOtherPage
}

if (originType === 'unrestricted-farm' && destination === 'afu') {
  return checkExistingLicenceExitPage
}
```

mukeshrawat02 (dismissed):

mukeshrawat02 (approved):

PR #800, opened 2026-05-27
Updates to follow latest Software Development Standards

joseluisgraa:


mukeshrawat02:
I think we need to update [.nvmrc](https://github.com/DEFRA/apha-apps-perms-move-animal-ui/blob/main/.nvmrc) and  [package.json](https://github.com/DEFRA/apha-apps-perms-move-animal-ui/blob/main/package.json#L9)  as well to  an active LTS version which is v24? 

Also, do we need to check with the DEFRA Standards team whether using TypeScript solely for JavaScript type-checking requires an exemption?
We currently have a [TypeScript configuration](https://github.com/DEFRA/apha-apps-perms-move-animal-ui/blob/main/tsconfig.json#L3C4-L4C3) with allowJs and checkJs enabled to provide type-checking for JavaScript files. This is in relation to the DEFRA requirement: “Don’t use TypeScript without approved exemption"

I think we should also add these as a checklist item for similar PRs, so we remember to manually check them in the repo settings:
- [x] Dependency graph enabled
- [x] Dependabot alerts enabled
- [x] Secret scanning enabled
- [x] Code scanning enabled
- [x] Protected branch required status checks

mukeshrawat02 (approved):

PR #810, opened 2026-05-29
Fix brittle GOV.UK guidance link selector on fifty percent warning page

davidkelley:
The exact link-text selector required the anchor's own text to equal "GOV.UK Bovine TB guidance page (opens in new tab)", but the template renders "(opens in new tab)" outside the <a>, so the anchor text is only "GOV.UK Bovine TB guidance page". Use a partial match so the BrowserStack test reliably finds the link.

The recent dependency changes was causing the brittle selector to fail.

Here is the Browserstack run for this branch: https://github.com/DEFRA/apha-apps-perms-move-animal-ui/actions/runs/26646748163


mukeshrawat02 (approved):
LGTM

PR #823, opened 2026-06-15
Refactor key facts to use Application Model instead of the state and …

joseluisgraa:
…not include irrelevant data


mukeshrawat02 (approved):
LGTM

PR #826, opened 2026-06-17
Updated answer models to have displayText instead of html getter, moved html to parent answer-model

joseluisgraa:
Updated answer models to have displayText instead of html getter, moved html to parent answer-model

### DIFF
--- a/src/server/common/model/answer/answer-model.js
+++ b/src/server/common/model/answer/answer-model.js
@@ -1,4 +1,5 @@
 import { NotImplementedError } from '../../helpers/not-implemented-error.js'
+import { escapeHtml } from '../../helpers/escape-text.js'
 /** @import {AnswerErrors, AnswerValidationResult} from './validation.js' */
 /** @import {RawApplicationState} from '../state/state-manager.js' */
 
@@ -32,7 +33,7 @@ export class AnswerModel {
     return {
       type: this.type,
       value: this.toState(),
-      displayText: this.html.replace(/<br \/>/g, '\n')
+      displayText: this.displayText
     }
   }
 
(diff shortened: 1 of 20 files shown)


src/server/common/model/answer/radio-button/radio-button.js
```diff
@@ -83,16 +83,15 @@ export class RadioButtonAnswer extends AnswerModel {
...
   get value() {
     return this._data?.[this.config.payloadKey]
   }
 
-  get html() {
-    const value = this._data?.[this.config.payloadKey]
-
+  get displayText() {
+    const value = this.value
     return this.config.options[value]?.label ?? ''
```
mukeshrawat02, line 95:
```suggestion
  return this.config.options[this.value]?.label ?? ''
```

src/server/common/model/answer/answer-model.js
```diff
@@ -86,10 +87,17 @@ export class AnswerModel {
...
    */
-  get html() {
+  get displayText() {
     throw new NotImplementedError()
   }
 
+  /**
+   * @returns {string}
+   */
+  get html() {
```
mukeshrawat02, line 97: I can see we've removed or replaced `get html()` everywhere else, for example here: https://github.com/DEFRA/apha-apps-perms-move-animal-ui/pull/826/changes#diff-f8ba600aacfefbbec4516e6a8949bf918ffc5c3c25ee32e6c9ff4983cd418344L84. So just to double check, do we still need this, or should we remove it as well?

mukeshrawat02: Ahh from your PR description I can see you mentioned that `moved html to parent answer-model`

src/server/common/model/answer/radio-button/radio-button.js
joseluisgraa: trying not to change anything I'm not sure about, that's why in this case I've left the existing logic

src/server/common/model/answer/answer-model.js
joseluisgraa: correct, the logic is common now and sits with the parent answer model

mukeshrawat02 (approved):

PR #875, opened 2026-08-06
Removed tokenUrl configuration

mukeshrawat02:


hughfdjackson:
The apha-integration-bridge `/oauth2/token` endpoint is just a development convenience, not intended for production use.  

You need to configure the token url separately - and use the corresponding cognito endpoint for the environment.  I've sent you a link to the internal docs.

